PR #12, opened 2021-11-23
Keycloak

mmalfertheiner:
- Initialize keycloak
- Update token every minute (access token has an expiration time of 5 minutes)
- Save token, isAuthenticated & user in global vuex state


databrowser/src/components/header/AppHeader.vue
```diff
@@ -2,15 +2,17 @@
   <header>
     <router-link to="/" class="text-2xl">OPENDATAHUB!</router-link>
     <LanguageSwitcher></LanguageSwitcher>
+    <user-authentication />
```
alexlanz, line 5:
```suggestion
    <UserAuthentication />
```

I would suggest to use this naming convention, because it is closer/the same as the component name.

databrowser/src/domain/auth/UserAuthentication.vue
```diff
@@ -0,0 +1,65 @@
...
+  <div v-else>
+    <button @click="onLogin">Login</button>
+    <button @click="onRegister">Register</button>
+  </div>
+</template>
+
+<script lang="ts">
+import { computed } from 'vue';
+import { useStore } from 'vuex';
+import { keycloak } from './init';
```
alexlanz, line 14: What do you think about renaming the `init` file to `keycloak`? There we store all the Keycloak logic.

alexlanz (approved):

databrowser/src/main.ts
```diff
@@ -13,6 +14,7 @@ app.use(router);
 
 // Add axios and provide it as injectable property (see https://www.npmjs.com/package/vue-axios)
 app.use(VueAxios, axios);
+app.use(store);
```
gappc, line 17: Putting this declaration here splits the axios config into two pieces. This is not a big deal, we can reorder it in one of the next commits in order to be able to apply and close this PR.

gappc (approved):

PR #14, opened 2021-11-27
Feature: refactor databrowser as defined by convention

gappc:


databrowser/README.md
```diff
@@ -59,6 +59,12 @@ If you are in the root folder of this mono repository, change to the Databrowser
 cd databrowser
 ```
 
+Provide a `.env` file with necessary configurations. You can use the `.env.example` file as basis:
```
alexlanz, line 62: Here I would adjust a bit the readme standard from the boilerplate: https://github.com/noi-techpark/java-boilerplate#execute-with-docker

databrowser/src/pages/DatasetListPage.vue
```diff
@@ -21,7 +21,7 @@
             :key="col.title"
```
alexlanz, line 21: What do you think about moving this file into a sub directory? For example like `pages/datasets/List.vue`. Like this we don't have everything at the top level.

gappc: I agree that this file needs to be relocated. Given that the list, detail an raw views are shown on the same page (switchable by the tab-like buttons "Table view", "Detail view" and "Raw data"), I would opt to keep something like `DatasetPage.vue` in the `pages` folder and to move the different views either into the `components` folder or to modularize it like you suggested and put it somewhere beneath the pages folder, e.g. `pages/datasets/...`.

If it's ok with you, I would do that refactoring in a later stage, I'm already working on the next PR where it could be part of (that PR contains improved table rendering and pagination).

databrowser/README.md
gappc: Do you mean that we should change the sentence to the one below (from the linked resource)?

`Copy the file .env.example to .env and adjust the configuration parameters.`

alexlanz: Yes, I just know that they try to allign all readme files as much as possible. But this is just a minor point.

databrowser/src/pages/DatasetListPage.vue
alexlanz: For me it is always hard to find files, if the pages folder does not follow the same structure as the components page. That's why I proposed it. But we can do this refactoring also in another PR.

databrowser/README.md
gappc: I think that's a good argument. I changed the sentence to the one shown above.

databrowser/src/pages/DatasetListPage.vue
gappc: If there is no differing opinion I'm going to decompose the `DatasetListPage.vue` into several smaller components using the naming scheme you suggested. The change will be done on a subsequent PR.

mmalfertheiner (approved):

PR #19, opened 2021-12-01
Raw dataset table

stplasim:


databrowser/src/pages/DatasetRawPage.vue
```diff
@@ -0,0 +1,83 @@
...
+    const configEntry = apiConfigProvider(route.params.datasetType as string);
+    if (configEntry != null) {
+      fetchList(
+        configEntry.detailEndpoint.path,
+        route.params.datasetId as string
+      );
+    }
+    return {
+      apiResult,
+      configEntry,
```
mmalfertheiner, line 73: Do you need to return this here? It's not used outside of the setup function.

databrowser/src/pages/DatasetRawPage.vue
```diff
@@ -0,0 +1,83 @@
...
+            v-for="(entry, index) in Object.entries(apiResult.data?.data)"
+            v-else
+            :key="index"
+          >
+            <TableCell>{{ entry[0] }}</TableCell>
+            <TableCell class="whitespace-pre-wrap">
+              {{ getEntryValue(entry[1]) }}
+            </TableCell>
+          </TableRow>
+        </template>
```
mmalfertheiner, line 25: In order to be more flexible I would propose the following API:

```suggestion
        <TableHeader>
          <TableHeader>Key</TableHeader>
          <TableHeader>Value</TableHeader>
        </TableHeader>
        <TableBody>
          <TableLoadingRow v-if="apiResult.isFetching" />
          <TableRow
            v-for="(entry, index) in Object.entries(apiResult.data?.data)"
            v-else
            :key="index"
          >
            <TableCell>{{ entry[0] }}</TableCell>
            <TableCell class="whitespace-pre-wrap">
              {{ getEntryValue(entry[1]) }}
            </TableCell>
          </TableRow>
        </TableBody>
```

mmalfertheiner: This would have the benefit, that the table header is not required and you can do whatever you want.

databrowser/src/pages/DatasetRawPage.vue
```diff
@@ -0,0 +1,83 @@
+<template>
+  <ContentArea>
+    <div>
+      Dataset Detail Page (type = {{ $route.params.datasetType }}, ID =
+      {{ $route.params.datasetId }})
+    </div>
+    <div>
+      <DataTable>
```
mmalfertheiner, line 8: Also here I would be more general: 
```suggestion
      <TableCustom>
```

I did the same in the button. As we cannot have `Table` alone due to Vue limitations. If you can think of a better name great, but DataTable is too specific IMO.

mmalfertheiner (changes requested):

databrowser/src/pages/DatasetRawPage.vue
stplasim: Would you have an example of that? I don't quite understand how this API works.

mmalfertheiner (approved):

databrowser/src/components/table/TableCustom.vue
```diff
@@ -0,0 +1,9 @@
+<template>
+  <div className="overflow-x-auto">
+    <div class="overflow-hidden min-w-full rounded-xl border border-gray-400">
```
mmalfertheiner, line 3: Due to overflow-hidden, the mobile view does not scroll and on the x-axis. We need a another way to handle this.

mmalfertheiner: You could try to fix it with pseudo classes `first, last` element.

https://tailwindcss.com/docs/hover-focus-and-other-states#first-child

mmalfertheiner (changes requested):

mmalfertheiner (approved):

PR #21, opened 2021-12-03
Feature implement list view

gappc:


gappc:
@alexlanz @mmalfertheiner @stplasim: please take a look at the implementation of the list view

**please be aware of the missing styling and cell renderer implementations (e.g. images)**

Current features:

- load data from API
- synchronized URL with API query parameters
- supports both Tourism API pagination (with properties Items, TotalPages, TotalResults, CurrentPage) and array-result pagination for endpoints that return a plain array
- provides utilty functions `useUrlQueryRouter` (to work with a set of url query parameters) and `useUrlQueryParameter` (to bind a single url query parameter to a reactive value)

databrowser/src/domain/api/client.ts
```diff
@@ -1,23 +1,22 @@
-import { inject } from '@vue/runtime-core';
-import { AxiosInstance, AxiosResponse } from 'axios';
 import { useQuery } from 'vue-query';
-import { UseQueryReturnType } from 'vue-query/lib/vue/useBaseQuery';
+import { QueryFunction, QueryKey } from 'react-query/types/core';
```
mmalfertheiner, line 2: Why do we have `react-query` as a dependency?

mmalfertheiner (commented):
Sometimes i reach a state, where the pagination is not highlighting the current page anymore. Then it also starts to be very laggy.

This happens when: First navigation on the list screen between pages.

databrowser/src/domain/api/client.ts
gappc: The library `vue-query` relies on the type definitions of `react-query`, because it is a port of `react-query`. I think there are no real react dependencies involved

gappc: Regarding the pagination bug: well spotted and thank you for the good description to reproduce the error!

I played a little bit around and found out that the pagination starts to work again as soon as the page size changes. Very strange, I'm going to take a look at it.

mmalfertheiner: I see 🙈 Nasty shit!

PR #22, opened 2021-12-05
Implement custom select box

gappc:


databrowser/src/components/select/SelectCustom.vue
```diff
@@ -0,0 +1,52 @@
...
+
+<script lang="ts">
+import { defineComponent } from '@vue/runtime-core';
+
+export default defineComponent({
+  props: {
+    modelValue: {
+      required: false,
+      default: () => null,
+      type: null,
```
mmalfertheiner, line 28: What does type `null` mean? Shouldn't it be `String`?

databrowser/src/components/select/SelectCustom.vue
```diff
@@ -0,0 +1,52 @@
...
+    <slot></slot>
+  </select>
+</template>
+
+<script lang="ts">
+import { defineComponent } from '@vue/runtime-core';
+
+export default defineComponent({
+  props: {
+    modelValue: {
```
mmalfertheiner, line 25: Could we just name it `value`?

databrowser/src/components/select/SelectCustom.vue
gappc: `null` in this case means arbitrary type, because the value could be anything (of course, most of the time it will be a string, but could also be a number). We could explicitly state the types like `[String, Number, Object, ...]` or leave it as null, what do you say?

databrowser/src/components/select/SelectCustom.vue
gappc: If I'm not wrong this is the name-by-convention to attach a `v-model` attribute, it won't work with another property name

mmalfertheiner: thanks for the explanation. 👍

databrowser/src/components/select/SelectCustom.vue
mmalfertheiner: Oh I see! Then it's fine I guess.

mmalfertheiner (approved):

stplasim (approved):

PR #24, opened 2021-12-10
Add download section

stplasim:


stplasim:
Hi @alexlanz , you are welcome to have a look at the PR. Please note that the CSV feature does not work yet. I searched a bit but didn't find anything that would work in the first place and is still maintained in the second. 

The API docs state that you can select CSV as the content type and thus get the data from the api, unfortunately this is not implemented.

databrowser/src/components/download/DownloadSection.vue
```diff
@@ -0,0 +1,145 @@
+<template>
+  <div class="mx-auto max-w-5xl">
+    <div class="hidden lg:flex items-stretch space-x-4 h-full">
+      <span class="self-center">Retrieve data based on table settings</span>
```
alexlanz, line 4: The strings should be translated.

databrowser/src/components/download/DownloadSection.vue
```diff
@@ -0,0 +1,145 @@
...
+
+      <DownloadButtonCell
```
alexlanz, line 6: I would make a component here with another name like `DatasetDetails`. And inside this component you could make use of general components like `DownloadCell`, `DownloadTitle` and `DownloadButton`. In addition, like this you have to advantage that you can group logic into the component. For example the `downloadJson` function in that you use in the component below.

databrowser/src/pages/datasets/RawViewPage.vue
```diff
@@ -10,6 +10,12 @@
         <div v-if="apiResult.isFetching">Loading</div>
         <div v-else>
           <vue-json-pretty :data="apiResult.data?.data" :deep="3" show-length />
+          <div class="fixed bottom-0 left-0 py-5 w-full bg-white">
```
alexlanz, line 13: Can we make this also as a component so that we can reuse it also in the detail and table view?

RudiThoeni:
> Hi @alexlanz , you are welcome to have a look at the PR. Please note that the CSV feature does not work yet. I searched a bit but didn't find anything that would work in the first place and is still maintained in the second.
> 
> The API docs state that you can select CSV as the content type and thus get the data from the api, unfortunately this is not implemented.

Hi Simon

The csv on the api should work see here https://github.com/noi-techpark/odh-docs/wiki/Export-data-to-csv-format
by adding a parameter https://tourism.opendatahub.bz.it/v1/ODHActivityPoi?format=csv
OR by passing the Header Accept: "text/csv"

BUT :) The question is what sense does it make to get this complex object with x Object Arrays etc. inside as csv, because CSV is more for a flat datastructure. Normally i used csv export by selecting single fields at example this call https://tourism.opendatahub.bz.it/v1/ODHActivityPoi?fields=Id,Detail.de.Title,ODHTags.[*].Id&format=csv

I think for the first mvp it is enough to add the possibility to get the whole data as csv, for the next iterations we have to  discuss it further, i think it makes only sense with field selection

alexlanz:
Thanks for the feedback @RudiThoeni. This is a point that I wanted to discuss with you in todays meeting. There we have to see if it really makes sence to provide a CSV download button.

RudiThoeni:
Yes, i think for MVP1 a button with a full csv export is enough, for MVP2 we have to think about a single field select concept, because this is what makes the csv export a useful functionality.....
Lets see what the PMs think about it

alexlanz:
A single field select concept? Do you have any use cases or examples where this could be useful for users?

RudiThoeni:
Yes at example Antonietta from IDM uses the CSV Export very often. She filters the data on their needs, then she only exports at example for a ODHActivityPoi the Detail.de.Title, LocationInfo.MunicipalityInfo.de.Name , ContactInfos.de.Email.
The other fields are not for interest for her.....So she gets a readable CSV with the data she is looking for....

the api has the "fields" filter which allows this single selection of the fields......
but this is all stuff for MVP 2 ;)

databrowser/src/components/download/DownloadJson.vue
```diff
@@ -0,0 +1,42 @@
...
+      type: String,
+      default: 'dataset.json',
+    },
+  },
+  setup(props) {
+    const url = computed(() => {
+      return `data:text/json;charset=utf-8,${encodeURIComponent(
+        JSON.stringify(props.dataset)
+      )}`;
+    });
```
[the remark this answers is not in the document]
stplasim: I have tried a bit but haven't found a way to do this directly with the API.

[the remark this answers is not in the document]
alexlanz: Ok, I will as Rudi. Thanks in the meantime.

databrowser/src/components/download/DatasetDetails.vue
```diff
@@ -0,0 +1,62 @@
...
+      <DownloadTitle>{{ $t('datasets.download.copyTitle') }}</DownloadTitle>
+      <DownloadDescription>{{ datasetUrl }}</DownloadDescription>
+    </div>
+    <DownloadButton
+      title="Copy dataset url to the clipboard"
+      @clicked="copyToClipboard"
+    >
+      <IconCheck v-if="showOk" />
+      <IconCopy v-else />
+      <span class="sr-only">Copy dataset url to the clipboard</span>
```
alexlanz, line 13: Text should be translated.

databrowser/src/components/download/DatasetDetails.vue
```diff
@@ -0,0 +1,62 @@
+<template>
+  <DownloadCell class="w-full">
+    <div class="inline-flex flex-col justify-center">
+      <DownloadTitle>{{ $t('datasets.download.copyTitle') }}</DownloadTitle>
+      <DownloadDescription>{{ datasetUrl }}</DownloadDescription>
+    </div>
+    <DownloadButton
+      title="Copy dataset url to the clipboard"
```
alexlanz, line 8: This is not required if you add the span below.

databrowser/src/components/download/DatasetDetails.vue
alexlanz: I think "Copy to clipboard" is enough.

databrowser/src/components/download/DownloadCSV.vue
```diff
@@ -0,0 +1,38 @@
+<template>
+  <DownloadCell>
+    <DownloadTitle>CSV</DownloadTitle>
+    <DownloadLink :download="filename" :url="url" title="Download CSV">
```
alexlanz, line 4: Same here. Title can be left away.

databrowser/src/components/download/DownloadCSV.vue
```diff
@@ -0,0 +1,38 @@
...
+      <IconDownload />
+      <span class="sr-only">Download dataset as CSV</span>
```
alexlanz, line 6: Strings should be translated.

alexlanz (approved):

PR #30, opened 2021-12-13
Dataset navigation

alexlanz:


gappc (approved):
Please add `<script lang="ts">` to the Vue files in order to improve tooling support. The rest looks good.

PR #32, opened 2021-12-16
Rewrite LanguagePicker with the new composition api

stplasim:


databrowser/src/components/language/LanguagePicker.vue
```diff
@@ -3,7 +3,7 @@
     class="inline-flex md:hidden items-center"
     @click="showMobileSelect = true"
   >
-    <span class="pr-2">{{ currentSelected }}</span>
+    <span class="pr-2">{{ currentSelected.toUpperCase() }}</span>
```
alexlanz, line 6: When a user reads this with a screen reader then the screen reader reads him for example "DE". I would add here an alternative description text for it. You can take a look at the Bookiply lanugage picker on desktop.

databrowser/src/components/language/LanguagePicker.vue
```diff
@@ -63,45 +64,55 @@ export default defineComponent({
...
+    }
+
+    function changeLanguage(language: string) {
+      languageParameter.value = language;
+      closeDialog();
+    }
+
+    function isSelected(current: string) {
+      return currentSelected.value == current;
+    }
```
alexlanz, line 105: Here you are using this method to determine if a language is selected. The problem here with Vue is that there does not really exist a strategy to calculate the state of every single language with a computed property.

A possible solution is just to extract the language into a dedicated component where you can caluclate the selected state with an computed property. This is a performance boost. That's why I would extract this into a dedicated component.

stplasim:
Hi @alexlanz, I've changed the PillGroup component. I just don't quite know how we should do this with the mobile variant. We use there the PillButton. 
We could make a wrapper for the PillButton but how do we call it?

Any suggestions

databrowser/src/components/pill/PillGroup.vue
```diff
@@ -1,44 +1,24 @@
 <template>
   <div>
-    <div
+    <PillGroupItem
```
alexlanz, line 3: What do you think about renaming this component to `PillButtonGroup` and `PillButtonGroupItem`? So it is more consistent with the link variant.

databrowser/src/components/pill/PillGroup.vue
```diff
@@ -1,44 +1,24 @@
...
       v-for="(item, index) in data"
       :key="item"
```
alexlanz, line 5: What if two values in the array are the same? For example ["2", "2", "1"]. Then the key would be the same. Would it maybe make more sense to use the index for the key? Or do we have there also some problems? For example if the pagination changes and then page 4 moves from index 3 to index 2. I'm not even sure by myself what to choose here.

databrowser/src/components/pill/PillGroupItem.vue
```diff
@@ -0,0 +1,66 @@
...
+    },
+    isLast: {
+      type: Boolean,
+      default: false,
+    },
+    item: {
+      type: String,
+      required: true,
+    },
+    currentItem: {
```
alexlanz, line 50: Here I would replace this value with a `isSelected` value.

databrowser/src/components/pill/PillGroup.vue
```diff
@@ -60,18 +40,13 @@ export default defineComponent({
       }
     );
 
-    function isSelected(current: string) {
-      return selected.value == current;
-    }
-
     function changeSelectedItem(item: string) {
       selected.value = item;
       context.emit('selectedChange', selected.value);
```
alexlanz, line 45: Also here I think it could be dangerous to use the value to return the selection. I would use the index, because it is more reliable.

alexlanz: I would switch the selection state completely to the index and not to the value.

databrowser/src/components/pill/PillLinkGroupItem.vue
```diff
@@ -0,0 +1,69 @@
...
+    </router-link>
+  </div>
+</template>
+
+<script lang="ts">
+import { defineComponent, PropType } from '@vue/runtime-core';
+import { computed } from 'vue';
+import { useRoute } from 'vue-router';
+
+export type PillLinkGroupType = {
```
alexlanz, line 37: In the code we did not suffix types with the word `Type`. That's why I would suggest to use another word here. For example: `PillLinkGroupItem`.

databrowser/src/components/pill/PillLinkGroupItem.vue
```diff
@@ -0,0 +1,69 @@
...
+  label: string | number;
```
alexlanz, line 38: The label can be just a string or not?

alexlanz (approved):

PR #37, opened 2021-12-20
Use cell postfix instead of renderer

gappc:


alexlanz (approved):
Wow, it looks really nice! 💯

PR #42, opened 2021-12-26
Extract TableContent RAW and edit buttons to dedicated components

gappc:


databrowser/src/domain/datasets/tableView/LinkDetail.vue
```diff
@@ -0,0 +1,32 @@
...
+      justify-center
+      items-center
+      m-1
+      w-8
+      h-8
+      hover:bg-gray-300
+      rounded
+      border
+    "
+  >
```
alexlanz, line 18: If you take a look at this button and the other button, then there is no difference in the styling. I would suggest to make one general `TableButtonLink` or `DetailsLink` component or something. And just pass then in the `TableContent` component the content itself into the component using a slot.

```vue
<DetailsLink :to="..."><EyeDetailGreen /></LinkDetail>
<DetailsLink :to="...">RAW</LinkDetail>
```

And I would also localize the string "RAW". In addition, for the details button there is no description and so it is hard for users with screen readers to know what is going on there.

gappc: I see what you mean but there is indeed a difference in styling, namely `text-xs text-green-500` applied to the RAW link. That's the reason I put them into separate components. I'm therefore not sure if we should combine them into one component.

Regarding localization and description: very good input, I will change the code accordingly

gappc: I would not do that, because then we have those classes applied, no matter what. For example, lets suppose there will be a `DetailsLink` labeled "Edit" with yellow text-color - this would already be problematic.

Instead we could define the base classes like in `LinkDetail` and provide additional classes like `text-xs text-green-500`  when we use the component. The implementation can be seen in https://github.com/noi-techpark/it.bz.opendatahub.databrowser/pull/42/commits/383fca383f2ac73e29e437c18a7d85ba54fa7b0f

alexlanz (approved):

PR #44, opened 2021-12-26
Configurable dataset hero description

gappc:


databrowser/src/domain/datasets/DatasetHero.vue
```diff
@@ -1,21 +1,22 @@
...
 
-export default defineComponent({
-  components: { HeroContainer, InternalLink, HeroSubTitle, HeroTitle },
-});
+const props = defineProps<{
+  title?: string;
+  subtitle?: string;
+}>();
+
+const { title, subtitle } = toRefs(props);
```
alexlanz, line 21: Uh, I didn't know that this is also working! I really like it. But should we dicuss if we should stick to this style in all components or switch to the other setup method that we used until now?

alexlanz (approved):

databrowser/src/domain/datasets/DatasetHero.vue
gappc: I was also not aware of that until recently and I have to say that I find this style very pleasant to work with, because it forces you to use the composition API and think therefore accordingly. I agree that we should discuss if we move to this style.

alexlanz: I saw also, that the handling of the component imports is way nicer. I like it a lot. We should really dicuss if we switch to this for all components.

PR #46, opened 2022-01-10
Add download section to the dataset detail page

stplasim:
Add the download section to the detail page.

Also cleanup data fetching to reduce code duplication.


alexlanz (approved):
LGTM. Please just make sure that we here don't have too much collisions with the changes from @gappc.

PR #48, opened 2022-01-12
Add dataset menu

stplasim:


databrowser/src/layouts/menu/DatasetMenu.vue
```diff
@@ -0,0 +1,102 @@
+<template>
```
alexlanz, line 1: I would not call this component `DatasetMenu`, because it can also be a menu with something else. The underlying view logic should be decoupled from the content that is shown. The datasets should just be the content.

databrowser/src/layouts/AppHeader.vue
```diff
@@ -1,12 +1,15 @@
...
+    <div class="flex items-center space-x-16">
+      <router-link to="/">
+        <img
+          :alt="$t('header.logo')"
+          class="logo"
+          src="/logo-open-data-hub-black.png"
+        />
+      </router-link>
+      <MenuPopover />
+    </div>
```
alexlanz, line 12: On mobile the menu button and the logo are switches around.

Bildschirmfoto 2022-01-12 um 13 18 17

databrowser/src/layouts/menu/MenuPopover.vue
```diff
@@ -0,0 +1,23 @@
...
+  <Popover class="relative">
+    <PopoverButton class="flex space-x-4">
+      <IconMenu />
+      <span>Menu</span>
+    </PopoverButton>
+    <PopoverPanel
+      class="inline-flex absolute z-10 py-8 px-2 bg-white divide-x shadow"
+    >
+      <DatasetMenu />
+    </PopoverPanel>
```
alexlanz, line 11: On mobile the panel is full screen and only one section is always shown.

Bildschirmfoto 2022-01-12 um 13 19 41

alexlanz: On desktop the panel is also full screen and the different sections are shown side by side.

Bildschirmfoto 2022-01-12 um 13 21 08

databrowser/src/layouts/menu/DatasetMenuSubList.vue
```diff
@@ -0,0 +1,79 @@
...
+          <span class="pr-24">{{ item.label }}</span>
+          <ArrowRight />
+        </button>
+        <router-link
+          v-if="item.url"
+          :to="item.url"
+          class="flex-1 py-2 pr-24 pl-2 w-full text-left"
+          >{{ item.label }}
+        </router-link>
+      </li>
```
alexlanz, line 39: The background color must go all the way from side to side.

Bildschirmfoto 2022-01-12 um 13 21 53

databrowser/src/layouts/menu/ContentMenu.vue
```diff
@@ -0,0 +1,146 @@
+<template>
```
alexlanz, line 1: What you think about this name for the component: `OverlayMenu`?

alexlanz: And then the list component could be called `OveralyMenuList`.

databrowser/src/layouts/menu/ContentMenu.vue
```diff
@@ -0,0 +1,146 @@
...
+        bg-white
+      "
+    >
+      <button class="block py-4 px-4" @click="closeDialog">
+        <IconClose />
+      </button>
+      <div class="inline-flex flex-row h-full divide-x-2">
+        <ContentMenuSubList
+          :items="content"
+          :title="$t('header.menu.allDatasets')"
```
alexlanz, line 35: I would not make this value hard coded. This could also come from the configuration object passed to this object. This means that the passed data is a `MenuItem` rather than an array of `MenuItem`.

databrowser/src/layouts/menu/OveralyMenuList.vue
```diff
@@ -0,0 +1,108 @@
...
+import ArrowRight from '../../components/svg/ArrowRight.vue';
+import { ref, watch } from 'vue';
+import ArrowLeft from '../../components/svg/ArrowLeft.vue';
+
+export type MenuLink = {
+  label: string;
+  url: string;
+};
+
+export type MenuCategory = {
```
alexlanz, line 62: What do you think about calling this `MenuColumn`? And the field `categories` could be called `items`, because there can also be links not just categories. What do you think?

databrowser/src/layouts/menu/OverlayMenu.vue
```diff
@@ -0,0 +1,131 @@
...
+        px-4
+        h-screen
+        bg-white
+        border-t
+      "
+    >
+      <OveralyMenuList
+        :item="lastElement"
+        :show-arrow="navigation.length !== 1"
+        @select-category="(menu) => navigation.push(menu)"
```
alexlanz, line 69: Just a question here because I'm curious. Is it here not also possible to use the `addSubMenu` function?

stplasim:
> Wow, what a nice solution! Nice job, I really like your approach 🔥
> 
> There are just some minor UI things that don't work nicely on Firefox:
> 
> Bildschirmfoto 2022-01-14 um 08 59 02

Have you installed Tailwind 3 @alexlanz

databrowser/src/layouts/menu/OverlayMenu.vue
stplasim: Since addSubMenu also has an index and if this is not available in the mobile view.

I have now simply changed the method a bit so that the index is optional

alexlanz:
Uh, that could be the problem, sorry. Now it works. But another thing that now does not work for me is the border.

Bildschirmfoto 2022-01-17 um 08 07 16

databrowser/src/layouts/menu/OverlayMenu.vue
alexlanz: Sounds good. It was just a quick question 😉

alexlanz (commented):
Clicking on the background of the overlay does not close the menu.

stplasim:
Should be fixed now @alexlanz

PR #51, opened 2022-01-14
update download section link

stplasim:


databrowser/src/components/pill/linkUtils.ts
```diff
@@ -0,0 +1,8 @@
+export function buildUrlWithQuery(url: string, queryParams: object): string {
```
alexlanz, line 1: Why is this utils function placed inside the `pill` folder? Isn't it a generic function?

alexlanz (approved):

databrowser/src/components/pill/linkUtils.ts
stplasim: Yeah what do you think if we move it into src/domain/datasets? Or where would you put it.

alexlanz: Is it related to datasets? I would rather put it somewhere in a shared folder or?

alexlanz (approved):

PR #55, opened 2022-01-17
Improve data fetching

gappc:


alexlanz:
@gappc could you please inform @stplasim when you are done with this PR and when everything is merged. Because he extracted the component that pops up from the bottom on mobile (see #58). So then he can finish his PR.

alexlanz (approved):

PR #58, opened 2022-01-17
Extract bottom sheet into a new component

stplasim:
Extract the the bottom sheet from the language picker and put in into a new reusable component.


alexlanz:
Why did you not use it already somewhere? You just extracted it.

stplasim:
At the moment we only need it in the language picker and I want to avoid conflicts with #55

alexlanz:
Ok, so I would leave it open until the other PR is done. And then go on with this one.

alexlanz (approved):

PR #59, opened 2022-01-18
Add AsyncView component

stplasim:
Add a general purpose async view component to handle loading, success and error state of a given request


alexlanz (approved):
Nice, I really like it.

stplasim:
Hi @alexlanz, maybe you can help me here. We use unknown since we don't know what type is being passed. The problem is that typescript wants you to do a type check at some point. 

We can make data and error generic, unfortunately it seems like that doesn't work in Vue and I don't want to make the two any either.

Do you have any idea?

We could also customize the AsyncView so that it uses the Axios types.

Translated with www.DeepL.com/Translator (free version)

alexlanz:
What is the type that the two properties could have?

stplasim:
data: `AxiosResponse<any, any>`
error: `Error | AxiosError`

alexlanz:
For now this should fix the problem. But I can see your concerns. One nice optimization could be to adjust the `AsyncView` component to use the render function: https://v3.vuejs.org/guide/render-function.html

And another thing that I don't like is that the data and error prop that is passed to the slots are not typed. This for example requires the check if the error is not undefined or null where the error is rendered. I did not find anything how this could be solved so that the slot props are also typed. Maybe you can find something?

If not, then I would leave it like this for now.

PR #61, opened 2022-01-21
Feature implement filtering

gappc:


databrowser/src/locales/en.json
```diff
@@ -40,5 +40,11 @@
       "downloadJSON": "Download dataset as JSON",
       "downloadCSV": "Download dataset as CSV"
     }
+  },
+  "components": {
+    "cancelSave": {
```
alexlanz, line 45: This seems to me a bit a strange object. Can we remove it and move the two fields just under `components`?

databrowser/src/domain/cellComponents/components/filters/CancelSaveButtons.vue
```diff
@@ -0,0 +1,21 @@
+<template>
```
alexlanz, line 1: We normally name these buttons the `ControllButtons`. I don't know if you like the name

alexlanz (approved):
Nice job!

databrowser/src/locales/en.json
gappc: Done with 0713945

databrowser/src/domain/cellComponents/components/filters/CancelSaveButtons.vue
gappc: Done with 295d9b6

PR #63, opened 2022-01-23
Enhance filters

gappc:


alexlanz (commented):
Sorry, but this PR is really hard to review for me. You are mixing here a lot of things into one single PR. One thing is the enhancement of the filter itself. And the other thing is a reorganization of files.

Especially the regoranization of the files seems really strange to me. I don't really understand why a couple of things are placed under to folder `src/domain/api/client`? What does this have to do with the client? And also the things under `src/domain/api/service` are not really clear to me. And the last point: what do the things under `src/domain/api/service/utils.ts` have to do with the API? Are these not just simple helper function?

Maybe we can discuss these things in a meeting to be quicker with the review.

gappc:
@alexlanz You are right, this pull request contains different things. The reason was, that I started with the filter enhancements. Those necessitated improvements to the API, where some helper functions were implemented. Those helper functions first were distributed across `lib` and `domain` folders but all in all they belong together into the same domain. Therefore the creation of API client and service.

Now to your questions:
- `api/client` contains the client that fetches data from the server, related functionality is also located here. It is the communication related part of `api`
- the `api/service` implements the parameter handling for the API together with the URL state management. The handling is decoupled from the client implementation to improve the separation of concerns. It has no knowledge of how data is fetched / mapped
- last, the `api/service` provides utility/helper functions that are reused in different places. In my opinion they should be part of the `api/service` implementation, because they are strongly coupled to it. Moving those utility functions to somewhere else (other domain?) would of course be possible but that just adds one more indirection without gaining anything else - at least at this stage of the project

PR #64, opened 2022-01-23
Filter improvements, config movements, SFC setup cleanup

gappc:


alexlanz (commented):
Also here you have a lot of changes mixed together. All these three points could easily life in a single PR. Like this it is taff for me to review. Can we discuss it in a meeting together?

gappc:
@alexlanz I forgot to mention in the description of this PR that it builds on PR #63, therefore all changes related to PR #63 are part of this PR. I don't know any better way to build on a yet-not-merged change, maybe do you have an idea?

If you leave the PR #63 related stuff aside and focus on commits starting from 4e0d467, then this PR should hopefully make a lot more sense as it does just one thing: switch to SFC setup style for all components that did yet not use that style.

alexlanz:
Ok, so then let's discuss first the PR #63 and then we will see. 🙂

PR #131, opened 2022-08-30
Add Event DetailView and small fixes

BlueRyse:
Webcams small fixes include:
 - Attribute naming issue;
 - New Tags section.

### DIFF
--- a/databrowser/src/config/tourism/event/event.detailView.ts
+++ b/databrowser/src/config/tourism/event/event.detailView.ts
@@ -15,37 +15,76 @@ export const eventDetailView: DetailViewConfig = {
               component: CellComponent.StringCell,
               fields: { text: 'Shortname' },
             },
-          ],
-        },
-        {
-          name: 'IDs',
-          properties: [
             {
-              title: 'ID',
-              component: CellComponent.StringCell,
-              fields: { text: 'Id' },
-              class: 'break-all',
+              title: 'Logo',
+              component: CellComponent.ImageCell,
+              fields: {
+                src: 'ContactInfos.{language}.LogoUrl',
+              },
+            },
+            {
+              title: 'Image',
+              component: CellComponent.ImageCell,
+              class: 'w-40',
+              fields: {
+                src: 'ImageGallery.0.ImageUrl',
+              },
             },
           ],
         },
         {
-          name: 'Data states',
+          name: 'Ids',
+          properties: [
+            {
+              title: 'Tv Info',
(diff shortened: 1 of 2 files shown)


gappc:
Hi @BlueRyse, thank you for your work.

It is a very useful component, but I think the implementation could be improved. Here is my feedback

### Bug in `href` computation

When you click on any link in the `WEBCAM URL` column you'll see that it doesn't work, because the URL is computed as `'//' + text`, where the value of `text` is provided to the component (see [here](https://github.com/noi-techpark/it.bz.opendatahub.databrowser/blob/d787d0a1323a8d3cb5460dadcfba3219449c56f4/databrowser/src/domain/cellComponents/components/cells/UrlCell/UrlCell.vue#L3)). If `text` is a full URL (e.g. `http://webcams.magnus.it/tvahrntal/webcam/rechts_MEGA_1.jpg`) the result is a `href` value of `//http://webcams.magnus.it/tvahrntal/webcam/rechts_MEGA_1.jpg`, which is not valid

There are several way to fix it, in my opinion it boils down to two directions:

1. stay simple and just use the provided value. This should suffice for URLs provided by ODH. @RudiThoeni: do you know if URLs provided by ODH are full in the sense that they have a scheme, domain, port and so on? (for reference see [https://developer.mozilla.org/en-US/docs/Learn/Common_questions/What_is_a_URL#basics_anatomy_of_a_url](https://developer.mozilla.org/en-US/docs/Learn/Common_questions/What_is_a_URL#basics_anatomy_of_a_url))

2. try to check if the provided URL is valid, otherwise add some handling like the addition of a scheme, domain, ... In that case we need to discuss the requirements of the algorithm

In my opinion 1. is the better choice if ODH provides full URLs. Otherwise we should start a discussion how to handle different URLs

### Improve configuration

It would be nice if the component could be further configured, e.g.

- separate link text from `href` (at the moment the link text has the same value as used for the `href` computation)
- possibility to specify a link target (like `_self`, `_blank`, ...). Maybe there could be a default value if the link target is not configured

@BlueRyse: what do you think about the suggestions?

lukmay:
Thank you, @gappc for pointing out this behavior.
For now, I deleted the section which produced the bug.
I think your improvement ideas are great! I will try to implement them and update you then.

gappc:
Hi @BlueRyse,

I saw that you fixed the bug, should we merge the component or do you want to add the proposed enhancements before the merge?

lukmay:
@BlueRyse I agree 100%

PR #136, opened 2022-09-09
District Detail View changes, new GPS Data implementation on Detail Views, new ExperienceArea Detail View and small fixes

BlueRyse:
@gappc With this PR I used another way to Implement the GPS Data in the Detail View. You can check it in the District DetailView config file. I think that in this way we don't need to develop a new component for GPS Data.

Honestly, it was an obvious one, and I don't know how we didn't see it before. Anyway, let me know if this solution is ideal. If so, I can start adding GPS Data in all the DetailViews config files.

### DIFF
--- a/databrowser/src/config/config.ts
+++ b/databrowser/src/config/config.ts
@@ -24,6 +24,7 @@ import {
   measuringPointConfig,
   snowReportConfig,
   eventConfig,
+  experienceAreaConfig,
 } from './tourism';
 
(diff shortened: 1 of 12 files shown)


gappc:
Hi @BlueRyse, I like your solution, I think we should add it to the configs. Please let me know when the PR is ready to merge

BlueRyse:
@gappc PR ready to be merged :)

PR #144, opened 2022-09-28
restyle active/inactive-badges

ila131:
Hello Christian,
I restyled the active/inactive-badges in agreement with Paul.

Thanks,

Ilaria

### DIFF
--- a/databrowser/src/domain/cellComponents/components/cells/stateCell/StateCell.vue
+++ b/databrowser/src/domain/cellComponents/components/cells/stateCell/StateCell.vue
@@ -19,37 +19,16 @@ withDefaults(
 );
 </script>
 
-<style>
+<style scoped>
+span {
+  @apply list-item list-inside py-1 px-2.5 rounded float-left;
+}
 .active {
-  @apply py-1 px-5;
   background-color: rgba(61, 149, 252, 0.1);
-  display: list-item;
-  list-style-position: inside;
-  padding-left: 10px;
-  padding-right: 10px;
-  padding-top: 2px;
-  padding-bottom: 4px;
-  padding: none;
-  float: left;
-  clear: left;
-  border-radius: 6px;
   color: #2264e5;
-  letter-spacing: 0.5px;
 }
 .not-active {
-  @apply py-1 px-5;
   background-color: rgba(255, 25, 87, 0.1);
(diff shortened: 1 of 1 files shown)


gappc:
Hi @ila131, thank you for your nice and concise PR. It is well done (tailored to one single thing) and the result looks nice, but the CSS styling can be improved by using more of the already available [tailwindcss](https://tailwindcss.com/) classes (see explanation at the end of this comment why tailwindcss) and by combining styles that apply to both active and inactive badges into one `span` declaration - no worries, CSS styling and working with CSS frameworks is not easy and needs a lot of practice ;)

The `style` part could be rewritten into something like the following:

```
<style scoped>
span {
  @apply list-item list-inside py-1 px-2.5 rounded float-left;
}
.active {
  background-color: rgba(61, 149, 252, 0.1);
  color: #2264e5;
}
.not-active {
  background-color: rgba(255, 25, 87, 0.1);
  color: #d12953;
}
</style>
```

Some explanation:
- the `style` element is enhanced with the `scoped` attribute. That's Vue's way to say that the CSS declarations should not leak outside of this component
- common style declarations for active and inactive badges (e.g. display as list-item, paddings, rounded borders, ...) are combined using the definitions for `span`. That way we don't have to repeat styles
- if you take a look at the `span` styles you'll see the [`@apply`](https://tailwindcss.com/docs/functions-and-directives#apply) declaration that is used to inline tailwindcss utility classes into the style definitions
- the only difference in active / inactive stylings is coloring

What do you think about the suggested changes?

**Why tailwindcss?**
- it covers most of the necessary stylings
- using it enhances styling consistency, e.g. `rounded` means "apply 4px rounded border ", which is easier to remember / reuse than writing the explicit styles everywhere it is needed
- it allows to adapt styles in a central place (change definition of `rounded` once and it is applied everywhere where `rounded` is used)
- it provides best practices out of the box and prevents some browser quirks
- since it uses a compilation step, it produces an almost minimal amount of CSS definitions

ila131:
Hey Christian,
thank you very much for your valuable comment. I think that using tailwindcss is way better than using css. In this way I avoid repeating styles. 

Thank you for your time in writing your explanations and suggestions :)

Ilaria

gappc (approved):

PR #146, opened 2022-09-29
New Width prop on ImageCell component to allow custom width

BlueRyse:
@gappc This is a new feature I made to allow us to pass a custom width when using the ImageCell component.

Despite this addtion.

Passing the width like this:
```
{
          name: 'Ski map',
          properties: [
            {
              title: 'Ski map',
              component: CellComponent.ImageCell,
              fields: {
                src: 'SkiAreaMapURL',
                width: "50%"
              },
            },
          ],
        },
```

doesn't make the image smaller, but changing the default value on the cell itself works.

Can you please provide some feedback to let it work even when passing a width?

Thanks.


gappc:
Hi @BlueRyse, it works as intended if you declare the `width` property as `params`, not as `field`. Declarations inside the `field` property are used to extract data from the dataset, declarations inside the `params` property are handed to the component 1:1.

Example config:

```javascript
{
  title: 'Image',
  component: CellComponent.ImageCell,
  class: 'w-40',
  fields: {
    src: 'ImageGallery.0.ImageUrl',
  },
  params: {
    width: '50%',
  },
},
```

Please check if this works as intended and let me know

BlueRyse:
@gappc It works as intended! Thanks!

If you don't have any feedback, I'd say we can merge this PR?

@ila131 @d3liaa you can use this feature to modify the width of the image for every case :)

gappc:
@BlueRyse, @ila131, @d3liaa: please note that the ODH image service is capable of scaling images when you add the query parameter `width` to the image url.

For example, if you add `width=200` as query parameter, the image will be scaled to a width of 200px (see http://service.suedtirol.info/imageresizer/ImageHandler.ashx?src=images/articles/rezeptartikel/4a06cfff-74e9-40f5-9f72-d98c1a2366c6.jpg&width=200)

There is a function that adds the width as query parameter to the URL (see https://github.com/noi-techpark/it.bz.opendatahub.databrowser/blob/3d8082483799f0519c7380391bebc435b1cff92f/databrowser/src/domain/image/image.ts).

The ImageCell component could be improved to support a resize value. The advantage would be, that the image is not downloaded in its full size and then scaled to the given width on the browser, but instead the resizing is done on the server (where it gets also cached) and only the resized image is downloaded => less download volume => usually it means that the images get loaded faster. Should we open a feature request (GitHub issue)?

PR #209, opened 2022-11-14
edit region + added contact to detail

d3liaa:
hei christian:)
i did the region edit view and added the contact section as i apparently forgot it sorry


gappc:
HI @d3liaa, thanks again :)

PR #210, opened 2022-11-14
Mun dis

d3liaa:


gappc:
@d3liaa again, nice work :)

Looking through your PR I saw that there are some usages of the [CellComponent.ToggleCell](https://github.com/noi-techpark/it.bz.opendatahub.databrowser/blob/development/databrowser/src/domain/cellComponents/components/cells/toggleCell/ToggleCell.vue) component (e.g. [in this config](https://github.com/noi-techpark/it.bz.opendatahub.databrowser/blob/8ab63a38725b3ffdf004fd2f8a325639a68e6ebf/databrowser/src/config/tourism/district/district.editView.ts#L80)).

Do you know maybe if there are values that can't be changed by the user and therefore the toggle should be non-changeable (disabled) for the user? If such cases exist, we could extend the `CellComponent.ToggleCell` component to support a "disabled" config property (or something like that).

d3liaa:
@gappc the toggle component that i know for sure should not be editable are for example [https://github.com/noi-techpark/it.bz.opendatahub.databrowser/blob/8ab63a38725b3ffdf004fd2f8a325639a68e6ebf/databrowser/src/config/tourism/district/district.editView.ts#L94](active on ODH) that i saw in the swagger marked as read only. But i simply left is as in the detail view , should i converted them in toggle cell with the disabled property? Is this property already available or were you talking about creating it?

gappc:
Hi @d3liaa, the ToggleCell component has now an option to be disabled (user can not change toggle state) as of commit https://github.com/noi-techpark/it.bz.opendatahub.databrowser/commit/8d5e8ce4fe593223fb16144cbee9a507a286fc71.

The example usage can be seen in the [district.editView.ts ](https://github.com/noi-techpark/it.bz.opendatahub.databrowser/blob/development/databrowser/src/config/tourism/district/district.editView.ts#L97) file.

PR #221, opened 2022-11-30
Revert Gps type being read only

BlueRyse:
As we decided, we will not follow the current data browser edit view configuration, so I reverted this change


gappc:
Hi @BlueRyse, the current change makes the GPS type editable. Is this intended? If so, is the PR ready for merge?

BlueRyse:
As far as I know yes, because we decided to make every Data that's not Read Only on the swagger editable (except for some cases, like IDs, and sources).

@sseppi This is correct, right?
Also, I would suggest to start a discussion regarding if we should implement all the GPS Data editable  (with single lines editable cells) in all the edit views, or if we should wait the component for it.

PR #319, opened 2023-03-14
Added coding guidelines

MatteoBiasi:


databrowser/guideline.md
```diff
@@ -0,0 +1,54 @@
...
+2. If the views define their respective components from the configuration file, they must use the ComponentRender component. To include the components with the ComponentRender component, they must be defined in /src/components/registerForComponentRender.js, and then the related components must be placed in the /src/components/ folder, using the following extension: "forRender". Eg, a component called Table that can ben rendered using the ComponentRender component, must be called Table.forRender.vue. Plus, all components built to be rendered with the ComponentRender, must respect the following predefined structure details:
+
+```
+1. If a components emits a data update, it must always call the emit event called "update" passing the data of the update as an object of type ForRenderComponentEmit >>> { prop: string, value: any } >>> is this correct?
+2. TODO?
```
gappc, line 50: I would keep this for now, we still have to find a good solution to define the interface for the rendered components.

One idea would be to provide some "hook" / function that covers the interface, something along the lines of (useWriteable.ts)[https://github.com/noi-techpark/it.bz.opendatahub.databrowser/blob/development/databrowser/src/domain/cellComponents/components/utils/writeable/useWriteable.ts] but more elaborated.

Example types:

```typescript
interface PropertyValue {
  prop: string;
  value: unknown;
}

type useRenderInterface = (attrs: Ref<Record<string, unknown>>) => {
  isWriteable: boolean;
  // ... other stuffy
  update: (updateData: PropertyValue | PropertyValue[]) => void;
};
```

Example usage:

```typescript
const { isWriteable, update } = useRenderInterface(...);
```

The advantage is, that the component implementer does not have to know how the renderer interacts with the component. That allows us to detach the interface implementation details from the usage and reduces the current pain point of attribute / event name pollution because names like "editable" and "update" could then be named anyway we want ([see here for problematic naming](https://github.com/noi-techpark/it.bz.opendatahub.databrowser/blob/development/databrowser/src/components/componentRenderer/ComponentRenderer.vue#LL6-L7)). Note that arbitrary names for attributes / events could have been chosen already, the current state is a tradeoff between naming and understandability by component implementers.

@MatteoBiasi what is your opinion about this solution?

databrowser/guideline.md
```diff
@@ -0,0 +1,54 @@
...
+```
+
+3. The list of components included in registerForComponentRender.ts must be entered in alphabetical order.
+4. The configuration of a component that renders data should be taken from Open API whenever available. The edit component should be read only per default till Open API doesn't provide the "readonly: true".
```
gappc, line 54: I would leave that sentence away, because it is to much detail for a component implementer and anyway, where would a component get the OpenAPI data from? In my opinion, such things as configuration resolution should be part of the application (Data Browser), not of the components that render data.

gappc (commented):
Thank for the PR, it looks nice, I added some feedback

gappc:
Hey @MatteoBiasi, please tell me when you are ready to merge the PR

MatteoBiasi:
> Hey @MatteoBiasi, please tell me when you are ready to merge the PR

I suppose first we have to do tomorrow a final review together, then we'll be able to merge.

gappc:
@MatteoBiasi thank you for your work :+1:

PR #343, opened 2023-04-12
district

d3liaa:
i adjusted some minor things without touching the location field or adding the related content


gappc:
@d3liaa your PR touches a lot of images and the fonts, I'm pretty sure this is not intentional.

Please undo those changes and, as this doesn't happen the first time, please take better care when pushing PRs, at least take a look at them after pushing ;)

PR #355, opened 2023-04-19
implemented map in the gps

d3liaa:
Hi Christian,
Stefano asked me to add the map that FlashBeing made for the Quickview to the Gps in the shared view.

Have a nice evening,
Delia;)

### DIFF
--- a/databrowser/src/config/builder/tourism/gpsData.ts
+++ b/databrowser/src/config/builder/tourism/gpsData.ts
@@ -24,6 +24,18 @@ export const gpsDataCategory = (): DetailElements => ({
         },
       ],
     },
+    {
+      name: 'Map',
+      properties: [
+        {
+          title: '',
+          component: CellComponent.QuickViewMapView,
+          fields: {
+            gpsInfo: 'GpsInfo',
+          },
+        },
+      ],
+    },
   ],
 });
 


gappc:
@d3liaa thank you for the PR, unfortunately this does not work: the map is shown, but the location is ignored (e.g. http://localhost:3000/dataset/edit/tourism/v1/Accommodation/2657B7CBCB85380B253D2FBE28AF100E?language=de#gps-data).

The problem is with the mapping: QuickViewMapView.vue expects an array of `Latitude` and `Longitude` values, but you provide just the `GpsPoints` object. What do you think, how could we solve this?

d3liaa:
This is really weird because in the ODHActivityPOi it works fine. I also noticed that in the datasets where the mark is not visible, the map is shown only in the edit view and not in the detail view. I'll have a more complete look and see if there is a solution to this. Sorry I didn't catch it, I just assumed that if it worked on one dataset it should be the same for the others. 
I'll keep you updated :)

d3liaa:
Hei Christian, 
before the mapping used GpsInfo array that does not exist in every dataset. So i changed the mapping to be able to use the GpsPoint , the same used in the GpsData.ts and GpsPointCell and thus should work with every dataset. 

Have a nice day,
Delia:)

gappc:
@d3liaa looks good now, thank you :+1:

PR #362, opened 2023-04-28
Venue room

d3liaa:
Hi @gappc @sseppi @pkritzinger ,
I noticed that the Venue dataset has a Room Detail Array with useful info that is not displayed. Since the format is practically the same for the Room Booked in the Event Short dataset I created a similar table using that as a template. 
In the table, I displayed: Shortname, Indoor, and SquareMeters attributes since these seemed the most relevant and easy to show.
I inserted this table in the Main Data section after the IDs.

Screenshot 2023-04-28 125035

### DIFF
--- a/databrowser/src/config/tourism/venue/venue.sharedView.ts
+++ b/databrowser/src/config/tourism/venue/venue.sharedView.ts
@@ -1,3 +1,4 @@
+import { CellComponent } from '../../../domain/cellComponents/types';
 import {
   DetailViewConfig,
   EditViewConfig,
@@ -29,6 +30,24 @@ export const venueSharedView = (): DetailViewConfig | EditViewConfig => ({
         {
           name: 'IDs',
           properties: [idReadOnlyCell()],
+        },
+        {
+          name: '',
+          properties: [
+            {
+              title: 'Room Details',
+              component: CellComponent.EditRoomVenueCell,
+              listFields: {
+                pathToParent: 'RoomDetails',
+                fields: {
+                  Shortname: 'Shortname',
+                  Indoor: 'Indoor',
+                  SquareMeters: 'SquareMeters',
+                },
+                attributeName: 'roomVenue',
+              },
+            },
+          ],
         },
         dataStatesSubCategory(),
         sourceSubCategory(),
--- a/databrowser/src/config/tourism/venue/venue.sharedView.ts
+++ b/databrowser/src/config/tourism/venue/venue.sharedView.ts
@@ -43,6 +43,8 @@ export const venueSharedView = (): DetailViewConfig | EditViewConfig => ({
                   Shortname: 'Shortname',
                   Indoor: 'Indoor',
                   SquareMeters: 'SquareMeters',
+                  Capacity: 'VenueSetup.0.Capacity',
+                  SetupType: 'VenueSetup.0.VenueCode',
                 },
                 attributeName: 'roomVenue',
               },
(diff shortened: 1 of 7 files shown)


gappc:
Hi @d3liaa, impressive work, thank you :+1: 

Some feedback:
- the "Indoor" value is a boolean (see [here](https://github.com/noi-techpark/it.bz.opendatahub.databrowser/blob/c02826e8aaa7b94cff04f50d542c6849e122f910/databrowser/src/domain/cellComponents/components/cells/editRoomVenueCell/EditRoomVenueTab.vue#L22)). I think this value is better handled by a toggle (e.g `ToggleCell`)
- the rooms in venue contain other data, like features, detail text, images and many more (take a look at the image below and `VenueRoomDetails` type in [Swagger](https://api.tourism.testingmachine.eu/swagger/index.html#/Venue/SingleVenue)). @sseppi @pkritzinger @RudiThoeni should we display / add that data? Note that the "old" Data Browser seems also to handle only a subset of the available attritbutes (see [here](https://frontend.tourism.testingmachine.eu/ExternalDataSources/VenueList))

image

d3liaa:
Thank you!!
Then I will wait on your feedback on what to add to the table :)
(I just updated the indoor attribute to a toggle)

gappc:
@d3liaa thank you for the change regarding the toggle, but I meant that a toggle should be used in the `EditRoomVenueTab` component (see my link above), not in the `EditRoomVenueTable` like you did in the last commit ;) Please change it in the Tab component, too.

pkritzinger:
hi @gappc @sseppi @d3liaa @RudiThoeni 
I have to admit that I do not know which attributes are of relevance here but would suggest to go the "easy" way and show the same as were shown in the "old" databrowser.

Btw: thanks a lot for noticing @d3liaa

d3liaa:
Perfect thank you for the feedback i will fix it ASAP :)

d3liaa:
Hi,
after looking at the VenueRoomDetail schema the only attributes i could find (other than the ones I already added: shortname, indoor, squaremeter) are SetupType and Capacity. 
I will add them.

d3liaa:
Hi, @RudiThoeni i noticed that in the databrowser v1 the attributes of the venue room details are not editable. I am having trouble making the attributes `capacity `and `venueTypes `editable with the components available (we would have to create a custom one), do you think that it is necessary that thery are editable? or even shown at all?

Thank you for your time,
Delia:)

RudiThoeni:
Hi Delia

at the moment it is not needed to have it editable since this data is only imported, but in the future maybe...... showing it would be great....

cheers


Tech Transfer Digital

Rudi Thoeni

T +39 0471 066 673
***@***.******@***.***>

NOI Techpark Südtirol / Alto Adige
A.-Volta-Straße / Via A. Volta, 13A
I-39100 Bozen / Bolzano
www.noi.bz.it<http://www.noi.bz.it/>

________________________________
Von: Delia Mennitti ***@***.***>
Gesendet: Donnerstag, 11. Mai 2023 16:00
An: noi-techpark/it.bz.opendatahub.databrowser ***@***.***>
Cc: Rudolf Thoeni, NOI ***@***.***>; Mention ***@***.***>
Betreff: Re: [noi-techpark/it.bz.opendatahub.databrowser] Venue room (PR #362)


Hi, @RudiThoeni<https://github.com/RudiThoeni> i noticed that in the databrowser v1 the attributes of the venue room details are not editable. I am having trouble making the attributes capacity and venueTypes editable with the components available (we would have to create a custom one), do you think that it is necessary that thery are editable? or even shown at all?

Thank you for your time,
Delia:)

gappc:
Hi @d3liaa, thank you for your work, it's merged now :+1:

PR #366, opened 2023-05-08
feat: minor QuickView GUI optimizations

MatteoBiasi:


databrowser/src/components/quickview/QuickViewCardOverviewContentText.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <p class="text-dialog">{{ value }}</p>
+  <p :class="{ 'text-dialog': true, underline }">{{ value }}</p>
```
gappc, line 2: I'm not sure what this is supposed to do, maybe this is a bug?

MatteoBiasi: Hi @gappc, this should be fine. Now the component supports an underline property which underlines the text if true.

MatteoBiasi:
> @MatteoBiasi thank you for the changes.
> 
> I have one comment (maybe it's a bug) and there are some compile time warnings (see https://github.com/noi-techpark/it.bz.opendatahub.databrowser/pull/366/files). Please take a look at them, thx.

All compile warnings have been fixed. 👍

gappc:
Hi @MatteoBiasi thank you for the fixes. There remain 4 lint warnings regarding Tailwind class names for opacity, e.g. "Classname 'hover:bg-opacity-90' should be replaced by an opacity suffix (eg. '/90')".

You can see them e.g. in the [Files changed](https://github.com/noi-techpark/it.bz.opendatahub.databrowser/pull/366/files) section of this PR. I can also confirm that the warnings appear when I run `npm run lint` using your branch on my local machine.

Could you please take a look at it?

databrowser/src/components/quickview/QuickViewCardOverviewContentText.vue
gappc: @MatteoBiasi I see, the underline property gave an odd impression but of course it's just `'underline': underline` - my bad ;) 

One improvement would be to apply the `text-dialog` class directly onto the class attribute like `class="text-dialog"`, given that it is always set. The `underline` property would stay as it is and be applied dynamically, like you show it here. While fixing the lint stuff below, it would be very nice if you could apply that change.

MatteoBiasi: I did the change. Usually I prefer this annotation to avoid to write the class attribute twice.

MatteoBiasi:
@gappc fixed. Strange that husky let me do the commit with those two warnings.

PR #387, opened 2023-06-07
Reuse pre-commit hook and information

henri-egger:
I added the pre-commit hook for husky with the behavior as discussed in the old PR, it will check for compliance before the npm lint as it is usually faster and only if the reuse-helper-tool is installed, if not it will only print a warning.

I also added documentation in README.md as well as in guideline.md


gappc:
@henri-egger thank you for your work.

Before merging, I think we need to clarify how to integrate reuse into the development and build process (see also discussion at the bottom of #386). It seems that the reuse-tool depends on either Linux, Python or Docker. We need to find ways to support those options or at least put some hint that those options exist at all and some information on how one could use Python / Docker for the pre-commit hook

henri-egger:
@gappc no need to merge yet until we haven't decided on what to do eventually, but just in the meantime i made the pre-commit-hook also work on windows via python.

gappc:
@henri-egger nice work, although I think that the pre-commit hook is not fixed for Windows but for Python (see also [comment on your other PR](https://github.com/noi-techpark/it.bz.opendatahub.databrowser/pull/390#issuecomment-1582935012))

.husky/reuse
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+if command -v reuse &> /dev/null
+then
+    (cd databrowser && reuse lint)
+elif command -v pipx &> /dev/null
+then
+    (cd databrowser && pipx run reuse lint)
+else
+    echo -e "\e[33mWarning: neither reuse nor pipx is installed, omitting reuse compliance check\e[0m"
```
gappc, line 10: I'd add a pointer / link to our reuse documentation. The idea is, that the warning message contains enough information on how to proceed to fix this "problem".

.husky/reuse
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+if command -v reuse &> /dev/null
+then
+    (cd databrowser && reuse lint)
+elif command -v pipx &> /dev/null
```
gappc, line 6: I know we had a discussion about Windows / Python and your solution looks good. But after rethinking it, I'm not sure if we should put the `pipx` check / invocation here.

The reason is, that this invocation would start to download / install stuff the first time without that a developer expects that to happen. Another problem could be, that the `pipx` invocation fails (e.g. the download fails because the user is behind a proxy). In that case the whole commit process may break.

I think it may be better to leave just the `reuse` check / invocation and to point the developer to some documentation that says what can be done to fix the "problem". For example for `pipx` invocation, one could define an alias / link / script to `pipx run reuse` with name `reuse` that can than be invoked by the pre-commit hook.

What is your opinion on this?

README.md
```diff
@@ -141,6 +141,14 @@ see the [tags on this repository](https://github.com/noi-techpark/it.bz.opendata
...
 The code in this project is licensed under the GNU AFFERO GENERAL PUBLIC LICENSE license, Version 3. See the LICENSE file for more information.
 
+### Reuse
+
+This project aims to be [REUSE](https://reuse.software) compliant. All files contained in the repository either have to be mentioned in the [dep5](.reuse/dep5) file or preferably contain a header containing SPDX copyright and licensing information.
+
+The project is checked for licensing and copyright information before every commit by a pre-commit hook if the [reuse-helper-tool](https://github.com/fsfe/reuse-tool) is installed as well as in the github CI which will fail if REUSE standards are not satisfied.
+
+For more information on how to make new files REUSE compliant check the project [guidline](databrowser/guideline.md).
+
```
gappc, line 151: Here we could add some problem resolution strategies with the pre-commit hook. But that is something we can also add in a future iteration.

gappc (commented):
@henri-egger thank you for this important PR! Please take a look at my comments and let me know what you think about them.

.husky/reuse
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+if command -v reuse &> /dev/null
+then
+    (cd databrowser && reuse lint)
```
gappc, line 5: `reuse` must be invoked on the root of the project in order for it to find the `.git` folder. We must not change to the `databrowser` folder before running `reuse`.

The line should just be `reuse lint`

databrowser/guideline.md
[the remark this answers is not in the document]
henri-egger: Reuse allows for multiple ways to declare the licensing information and copyright holder for files, so we decided on some guidelines [here](https://github.com/noi-techpark/odh-docs/wiki/REUSE). If things are done diffrently the project will still be compliant just not in the way we want it to be. Therefore i used should. I will also add a link to this wiki page, that way contributors have the full information.

.husky/reuse
henri-egger: Yes i noticed that too, interestingly the lint executes the exact same way, but I will change it anyways.

henri-egger:
@gappc I applied the changes you proposed, let me know what you think.

gappc:
@henri-egger very good work, thank you :+1:

Issue #345, opened 2023-04-13
File line endings: Improve project compatibility with Windows


gappc:
Line endings in Windows OS are different than those in *ix OS. This is a problem for GIT commits. 

One solution seems to be to set the line endings in `.gitattribute` but we have to check if this creates problems before we can use such settings.

henri-egger:
@gappc
Summarizing PR #390:
- Just adding a Prettier linting rule to ignore line endings would solve the problem for now but could lead to problems later.
- Some contributors working on Windows might already have found a local solution, would need to reach out to them.
- Adding a `.gitattributes` file containing `* text=auto eol=lf` to the project root seems to be best option, the official Prettier docs even mention this solution. However, this is intended to be implemented when creating a new project. When implementing it in an existing repository it seems that the file has to be on the branch which will be referenced in HEAD upon cloning, which is the reason it wasn't working for me the first time. This means that the file should take effect when cloning normally once put into the development branch. Right now this is testable with `git clone <origin> --branch windows-linting`.
  You might just want to create the file via Github if this solution is ok for you, or i could make a new branch without the accidental git-hook commits and open a PR.

gappc:
@henri-egger very nice overview, thank you :100: 

Please go on and create a PR with the suggested changes.

PR #390, opened 2023-06-08
Fix prettier linting for windows

henri-egger:
While adapting the husky pre-commit-hook for windows i noticed that it wouldn't even have been possible for any windows user to commit as prettier doesn't allow crlf end of lines.

I just added a rule to the .eslintrc.json file to allow for those kinds of line endings.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto eol=lf
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,4 +1,6 @@
 #!/usr/bin/env sh
 . "$(dirname -- "$0")/_/husky.sh"
 
+bash ./$(dirname -- "$0")/reuse
+
 (cd databrowser && npm run lint)
--- a/.husky/reuse
+++ b/.husky/reuse
@@ -1,8 +1,11 @@
 #!/bin/bash
 
-if ! command -v reuse &> /dev/null
+if command -v reuse &> /dev/null
 then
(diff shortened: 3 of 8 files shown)


henri-egger:
@gappc yes the reuse commits are put in by accident, i will provide a new clean PR once we have decided on a solution.

According to the official Prettier documentation, it would actually make more sense to create a .gitattributes file containing
`* text=auto eol=lf`. I tried it however, following the exact steps mentioned [here](https://prettier.io/docs/en/options.html) and couldn't get this solution to work.

You mentioned there are already Widows users working on the project, i believe however that they are using WSL as that is very common for Windows users doing web development. Also i tried doing a clean clone of the repo on Windows without wsl and i was not able to commit due to the prettier linting, so even if some people are already working on Windows which might have found a personal solution, new Windows contributors still have this problem.

gappc:
@henri-egger thank you for the feedback.

You mentioned, that the `.gitattributes` solution did not work, what was the problem? Reading through the link you provided, it seems that the line-ending problem should be solvable by these configurations. It would be great if you could take a second look at it.

If I may ask you another favor: could you please put your comments and current knowledge (problems, solutions, links) regarding the line-ending problem from this PR to issue #390 such that we can keep discussing there? I think that would be the appropriate place and the info won't get lost. Thx :)

henri-egger:
@gappc i will try to take a closer look and move the discussion, did you mean to link issue [#345](https://github.com/noi-techpark/it.bz.opendatahub.databrowser/issues/345) or would you like me to create a new issue #390?

Issue #204, opened 2022-11-08
As an Open data Hub manager I want a better and simpler navigation


gappc:
The current Data Browser menu entries contain all endpoints of the ODH tourism domain from ~ July 2022.
Some entries are not of interest for the normal user (e.g. ODHTags).
We need to define which entries the menu should contain.

The next tasks are:
- [x] @pkritzinger analyzes the actual navigation
- [x] @pkritzinger collects feedback from the first beta users
- [x] @pkritzinger reviews the menu entry and makes a new proposal
- [x] the gorup reviews and agrees the final proposal 
- [x] @gappc implements the new menu

d3liaa:
I wanted also to add.
The menu contains entries such as Activities, Activities and Points of Interest and  Points of Interest. 
From these only "Activities and Points of Interest" should exist as is the only one with a respective endpoint the others do not exist or are deprecated.

sseppi:
I added this issue the the Mileston 2.2 since it is strongly connected to the issue #278 which aims to improve the navigation of the entire data browser (e.g. header, footer, connection with other Open Data Hub tools, metadata API, etc.)

d3liaa:
I reviewed the menu and created a new proposal. Maybe if @pkritzinger has time he can review it and decide if it needs further modifications.

New menu proposal:

 Locations
	-   [District](https://databrowser.opendatahub.testingmachine.eu/dataset/table/tourism/v1/District)
	-   [Experience Area](https://databrowser.opendatahub.testingmachine.eu/dataset/table/tourism/v1/ExperienceArea)
	-   [MetaRegion](https://databrowser.opendatahub.testingmachine.eu/dataset/table/tourism/v1/MetaRegion)
	-   [Municipality](https://databrowser.opendatahub.testingmachine.eu/dataset/table/tourism/v1/Municipality)
	-   [Region](https://databrowser.opendatahub.testingmachine.eu/dataset/table/tourism/v1/Region)
	-   [Ski Area](https://databrowser.opendatahub.testingmachine.eu/dataset/table/tourism/v1/SkiArea)
	-   [Ski Region](https://databrowser.opendatahub.testingmachine.eu/dataset/table/tourism/v1/SkiRegion)
	-   [Tourism Association List](https://databrowser.opendatahub.testingmachine.eu/dataset/table/tourism/v1/TourismAssociation)
[...]

PR #398, opened 2023-06-19
Fix menu overlay

henri-egger:
While testing the databrowser i noticed that there was a (mainly) visual bug when opening the menu whilst in a dataset view. The backdrop doesn't cover the toolbox and parts of the language picker as seen here:

databrowser-overlay-glitch

This also allows the user to interact with these parts of the ui while having the menu open.

The glitch is caused by the toolbox and the language picker having a higher z-index than the menu backdrop, that being 30, i fixed it by increasing the backdrops z-index to 40. Even though 30 would have been enough, i think it's clearer this way.

Raising the backdrops z-index doesn't seem to cause any other problems, so here the PR for the fix.

### DIFF
--- a/databrowser/src/layouts/menu/OverlayMenu.vue
+++ b/databrowser/src/layouts/menu/OverlayMenu.vue
@@ -27,7 +27,7 @@ SPDX-License-Identifier: AGPL-3.0-or-later
 
   <Dialog
     :open="dialogOpen"
-    class="fixed top-0 z-20 w-screen p-0"
+    class="fixed top-0 z-40 w-screen p-0"
     @close="closeDialog"
   >
     <DialogOverlay


gappc:
@henri-egger thank you for bringing up this topic. I noticed  another z-index problem (language picker visible in mobile on opened menu, see screenshot below).

Given that the menu will change due to #204 I suggest to put the work on this PR on hold until it is clear what the new design will look like. Then we need to look into the z-index stuff in detail (maybe you have some time to help out?). At the moment the used z-index are a bit of a mess due to my sloppiness ;)

Screenshot_20230619_204130_Brave

henri-egger:
Again, no problem, I think however that you can close this PR as the issues will be fixed anyways.

I will see if I can implement some kind of centralized solution for z-index constants, since that will probably be useful later anyways.

Also, not on topic with this PR, but will the /dataset-overview page also recieve a rework? If not I may try finding a solution to virtualize the large list there for optimization purposes.

Issue #412, opened 2023-07-20
MetaData API add Tag functionality


sseppi:
As a Meta Data Data Editor I want to have the Tag functionality active in the Meta Data API, to be allowed to define the Domains related to each Dataset I input.

Tasks:
- [x] Add the Tags to the APIS
- [x] Add the list of the Domains (e.g. Tourism, Mobility, Environment, Energy, Agriculture, Weather, Traffic, etc.) to the tags
- [x] Configure the Tag component in the Data Browser

RudiThoeni:
Hi Stefano

At the moment in the databrowser we have a Webcomponent for managing ODHTags
The data is stored in a string Idlist and the List "ODHTags" is generated automatically by the api out of this list and provide Links to the Tag info etc...

I added this functionality to the MetaData api the field to store the tags is `OdhTagIds`
No need to reimplement anything we can use all what is there......


For the new Tags Functionality (which was introduced to test in ODHActivityPois) where the Tag source is also  saved we have to create a new Webcomponent, but i think here we have to do some alignment first on how we would proceed with the Tagging in ODH for all datasets......

sseppi:
@RudiThoeni thank you for the fast implementation and the explanation!

@d3liaa can you please configure this new field in the MetaData API so that we can test if everything s working as expected?

d3liaa:
@sseppi sure :) I'll start working on it right away!

d3liaa:
@RudiThoeni sorry to disturb you. But I cannot find the field `OdhTagIds` in the Meta Data.

RudiThoeni:
@d3liaa I added it into the data model
![image](https://github.com/noi-techpark/it.bz.opendatahub.databrowser/assets/30471361/0f719633-ad16-40c5-96af-3f1c2a3626e6)

 on the next save operation it will be also in the json....

d3liaa:
@RudiThoeni Ah thanks perfect i understand :)

PR #413, opened 2023-07-24
added tags in the metadata shared view

d3liaa:
Hi, I configured the tags field in the shared view of the Meta Data. The field in the JSON is not available yet so i have to make a last check :)


d3liaa:
As asked in Issue #412 i added the Tag functionality. I tested everything and it works as expected. It is ready to merge if @gappc and @sseppi agree. :)

PR #420, opened 2023-08-22
Feat/new filters menu in Datasets Overview page

MatteoBiasi:


gappc:
Hey @MatteoBiasi, thank you for this PR, looks good overall and i like it :+1: 

Please don't get scared because of the number of review comments (especially on `OverviewListPage.vue`), most are best-practices I wanted to share like typing.

I encountered just one strange behavior when I tried the feature: if I set filters of the same type, they seem to work in a logical `AND` fashion instead of the expected `OR`. In the image below, I expected to see the datasets that belong to `tourism` OR `mobility` when I select both Dataspace filters. Instead, the result is empty, which seems odd.

image

MatteoBiasi:
> Hey @MatteoBiasi, thank you for this PR, looks good overall and i like it 👍
> 
> Please don't get scared because of the number of review comments (especially on `OverviewListPage.vue`), most are best-practices I wanted to share like typing. Note, that I didn't mark all code parts, e.g. I didn't mark all code parts where types could be improved. Please check for your self where comments apply.
> 
> Testing the feature, it seems to run fine, I encountered just one strange behavior: if I set filters of the same type, they seem to work in a logical `AND` fashion instead of the expected `OR`. In the image below, I expected to see the datasets that belong to `tourism` OR `mobility` when I select both Dataspace filters. Instead, the result is empty, which seems odd.
> 
> image

Hi @gappc, thanks for the review, but I cannot see the comments. 😅 This is the only comment I see. Can you double check if you can see them?

For the second part of the message, I had the same doubt. At the moment the frontend always uses an AND and never an OR. This is more an UX issue than coding: by my point of view, it makes sense for some filters to act as an AND; for example for tags, you may want all Datasets which match 2 or more tags. But for some others, like the Dataspace, it totally doesn't make sense since the Dataspace can only be a or b, and not have both. We may generalize and say that all filters which filter on a parameter which is an Array, should use the AND logic, while those who match on a String, should use the OR logic. But then at the same time we need to improve the UI showing for example rounded checkboxes (which will remind the input radio) were it's only possibile to choose a value and keep the squared checkboxes where it's possibile to choose multiple values. This is something that definitively has to be discussed. I quote @mrabans so maybe we can discuss it in the next meeting.

gappc:
@MatteoBiasi strange, I can see them as long as I'm logged in (if I'm logged out I can't see them either). Maybe I have to finish the review, just a sec...

MatteoBiasi:
> @MatteoBiasi strange, I can see them as long as I'm logged in (if I'm logged out I can't see them either). Maybe I have to finish the review, just a sec...

@gappc yep that was the issue. Now I can see them. 😊

mrabans:
> This is something that definitively has to be discussed. I quote @mrabans so maybe we can discuss it in the next meeting.

@MatteoBiasi absolutely, bring the topic up on Tuesday, so we can discuss it.

databrowser/src/pages/datasets/overview/OverviewListPage.vue
```diff
@@ -6,50 +6,171 @@ SPDX-License-Identifier: AGPL-3.0-or-later
...
+            </div>
+          </div>
+
+          <!-- Filters list -->
+          <div class="h-full overflow-y-auto py-14 md:h-auto md:p-0">
+            <div
+              class="truncate border-t border-gray-300 px-3 py-2 text-dialog"
+              @click="toggleFilter('hasNoMetadata')"
+            >
+              <ToggleCustom v-model="_inputModels.no_metadata" class="mr-2" />
```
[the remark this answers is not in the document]
MatteoBiasi: From my perspective, using the array filters to also handle the v-model of the checkboxes would only further complicate the code and make it even less performant because it would dynamically need to search in the filters array to determine whether the checkbox value is true or not. I've optimized the management of the _inputModels array by creating a new method called setInputModelKeyValue, making everything more robust and maintainable. Regarding the comment below, for this optimization, I would also avoid refactoring the code to update _inputModels all at once, as it would make it less readable, and there is currently no performance concern.

MatteoBiasi:
@gappc All PR issues have been fixed (check my only comment), the filtering logic has been optimised handling and/or between macro filters groups and group options. Everything should be ready to me merged. 😊

databrowser/package.json
```diff
@@ -37,6 +37,7 @@
     "vue-router": "^4.2.2"
   },
   "devDependencies": {
+    "@types/lodash": "^4.14.197",
```
gappc, line 40: Please remove the lodash type import

databrowser/src/pages/datasets/overview/OverviewListPage.vue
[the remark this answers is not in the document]
gappc: @MatteoBiasi my main concern is about having contradicting states. First, a developer must take care to not mismatch the variables for `toggleFilter` and `v-model`. Second, by design there is always a short period of time, where the states mismatch, because the `click` handler and `v-model` changes cannot execute at the same time. This can lead to really nasty bugs that are hard to track (race-conditions).

My gut feeling tells me that the components state management should be revised (see also the other issue about `_inputModels` update). On the other hand, the feature works as expected and if one day we'll find out that we have to improve, then we should be able to do it without greater struggle. So let's keep the code as provided now.

databrowser/src/pages/datasets/overview/OverviewListPage.vue
```diff
@@ -59,21 +174,445 @@ SPDX-License-Identifier: AGPL-3.0-or-later
...
+const mapDatasetsKeyArrayToFilterItems = (
+  datasets: TourismMetaData[],
+  key: TourismMetaDataIndexes
+) => {
+  const addedProps: Record<string, boolean> = {};
+  let finalList = [] as FilterSelectItem[];
+
+  datasets.forEach((dataset) => {
+    if (dataset[key] !== undefined) {
+      (dataset[key]! as string[]).forEach((propName) => {
```
gappc, line 263: Exclamation mark can be removed, as you have checked for existence beforehand.

Minor point: you just check for `undefined` in the line above, what about `null` values? You can check both for `undefined` and `null` at once using the `!= null` check like this:

```js
if (dataset[key] != null) {
```

That's 95% of the time the kind of check that you want to do

gappc:
@MatteoBiasi thank you for the changes. The only blocker is the remaining type import for `lodash`, the other stuff is optional

databrowser/src/pages/datasets/overview/OverviewListPage.vue
MatteoBiasi: I thought strict equality was required. Refactored. 👍

MatteoBiasi:
@gappc the latest comments have been soled. Should be ready to merge. 👍

gappc:
@MatteoBiasi nice work, thank you :)

Issue #439, opened 2023-10-24
Difference between detail/edit view and Add a new record view


d3liaa:
The add a record configuration is different from the detail / edit view. 
The add a record view should be the same as the edit but only blank. 
I noticed this for example in the article dataset:
As is visible in the screenshots below there are two fields that should not be present in the Main Data and the published on is completely missing from the configuration.
![article-ss](https://github.com/noi-techpark/it.bz.opendatahub.databrowser/assets/101118017/3b2eb14e-e6ef-4ad6-bfdf-144b1744d1cd)
![article-ss1](https://github.com/noi-techpark/it.bz.opendatahub.databrowser/assets/101118017/d508ab71-55cb-43af-b9c0-bb8e9407458f)

@dulvui since you have to touch the code, can you please remove the beta badge hard-coded and apply it in a parametric manner. The goal is to have the following behaviour:
- databrowser.opendatahub.testingmachine.eu --> Testing badge
- v2-beta.databrowser.opendatahub.com --> Beta badge
- databrowser.opendatahub.com --> No badge

gappc:
@d3liaa @sseppi yes you are right with your observation.

I checked, and the only two datasets with dedicated configuration for add-a-new-record are [Articles](https://github.com/noi-techpark/it.bz.opendatahub.databrowser/blob/development/databrowser/src/config/tourism/article/article.newView.ts) and [PublishedOn](https://github.com/noi-techpark/it.bz.opendatahub.databrowser/blob/development/databrowser/src/config/tourism/publishedOn/publishedOn.newView.ts).

If @d3liaa 's assumption is correct (new view should be the same as edit / detail view), the fix is as follows:

- remove the files mentioned above
- update the respective [articles.views.ts](https://github.com/noi-techpark/it.bz.opendatahub.databrowser/blob/development/databrowser/src/config/tourism/publishedOn/publishedOn.views.ts) and [publishedOn.views.ts](https://github.com/noi-techpark/it.bz.opendatahub.databrowser/blob/development/databrowser/src/config/tourism/publishedOn/publishedOn.views.ts) files

This change would be a good starter / project-introduction issue if anyone is interested in.

sseppi:
@pkritzinger @Mazzolintis In my point of view the "New record" and the "Edit" should be the same for a Dataset. The only difference between the two is that the "New record" is completely empty.

If you don't see any contra, I would suggest to plan to do the changes proposed by Chrsitian.

[...]

PR #446, opened 2023-11-02
'Add record' is now the same as 'Edit' view for Article and PublishedOn

dulvui:
'Add record' is now the same as 'Edit' view for Article and PublishedOn #439


gappc (approved):
Hey @dulvui nice work thx 👍

PR #458, opened 2023-11-21
Optimizing Editviews of various Datatypes

RudiThoeni:
This PR contains Optimizations and fixes for editing/inserting of various Datatypes including
- ODHActivityPoi
- EventShort
- Accommodation


RudiThoeni:
@cgapp i had some troubles with de CRLF, LF line endings (old linux/windows problem), without this config after a `npm run lint -- --fix` all files appeared as changed and i had CRLF as line endings.....  By installing the VS Code prettier eslint extension it seems this problem is gone...... so i will remove this config in the next PR (see here https://github.com/noi-techpark/it.bz.opendatahub.databrowser/pull/469/commits/5719d439160c6e2184ec8ed0c407e5969ac35617)

gappc:
@RudiThoeni I've opened the issue #470 and have assigned it to me

RudiThoeni:
> @RudiThoeni good to know, maybe we should add this information to the developer docs. I'll give it a try and do a PR.

;) i found this config in this post here https://stackoverflow.com/questions/53516594/why-do-i-keep-getting-eslint-delete-cr-prettier-prettier

It is always a mess with this line endings on windows...

gappc:
@RudiThoeni yeah, line endings... I agree 120% with what you say ;)

PR #482, opened 2023-12-14
Implement support for datasets from mobility domain

gappc:
This single giant commit changes a lot of stuff. First and most important, it adds support for datasets from the mobility domain. During this process, a lot of code was refactored and improved. The most important other changes are:

- rename config property "fields" to "objectMapping" to help in understanding
- rename config property "listFields" to "arrayMapping" to help in understanding
- rework dataset config generation and computation to simplify the code
- rework navigation and pagination for the table view
- simplify data fetching and processing
- unify the way how cell components fetch data
- relocate files and folder inside the "domain" folder
- add / improve components
- fix bugs in existing of components

... and lots of other stuff

PR #484, opened 2023-12-15
TO DELETE: Feat/delete record from table view

### LINKED ISSUE #460: As an Open Data Hub Content Editor I want to be able to delete the content I'm responsible for, to remove wrong records.
We need to implement as soon as possible the delete functionality, since some user are asking for it.

One idea could be to add a button `Delete` in the right column in Table view.

Shall we add a the delete functionality also in the detail, edit and raw view?

a-crea:
This PR resolves #460

### DIFF
--- a/databrowser/src/components/dialog/DialogOverlay.vue
+++ b/databrowser/src/components/dialog/DialogOverlay.vue
@@ -9,3 +9,7 @@ SPDX-License-Identifier: AGPL-3.0-or-later
     <div class="fixed inset-0 bg-black/25"></div>
   </DialogOverlayTransitionChild>
 </template>
+
+<script setup lang="ts">
+import DialogOverlayTransitionChild from './DialogOverlayTransitionChild.vue';
+</script>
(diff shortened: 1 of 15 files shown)


a-crea:
@gappc is there a way to easily reload the current dataset?
Example: I'm in the event short table view, I delete a record and the table reloads. I managed to achieve that before the major merge of #482, but there should be an even easier and cleaner way now.

gappc:
Ah, I think that's related to [TanQuery](https://tanstack.com/query/latest) that is used to fetch data. Internally it uses a caching algorithm, keyed by the URL. I think the solution could be around here: https://tanstack.com/query/latest/docs/react/guides/query-invalidation

Could you please provide me the steps to reproduce the problem, maybe using your branch?

a-crea:
@gappc we can ignore and close this PR :) Please consider the #489

sseppi:
I close without mergin this issue, since a-crea created a new one for the same features.

PR #489, opened 2023-12-15
Feat/delete record from table view rebase

### LINKED ISSUE #460: As an Open Data Hub Content Editor I want to be able to delete the content I'm responsible for, to remove wrong records.

a-crea:
This PR resolves #460

### DIFF
--- a/databrowser/src/components/svg/IconCloseCircled.vue
+++ b/databrowser/src/components/svg/IconCloseCircled.vue
@@ -1,0 +1,28 @@
+<!--
+SPDX-FileCopyrightText: NOI Techpark <digital@noi.bz.it>
+
+SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+
+<template>
+  <svg viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg">
+    <path
+      d="M10.4229 5.57715L5.57715 10.4229"
+      stroke-width="1.14286"
+      stroke-linecap="round"
+      stroke-linejoin="round"
+    />
+    <path
+      d="M5.57715 5.57715L10.4229 10.4229"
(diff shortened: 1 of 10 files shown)


a-crea:
Hi @gappc , please let me know if everything is in order. I had to create new branches and rearrange the code, so now everything should proceed smoothly :)

gappc:
Hey @a-crea, impressive work, well done :+1: 

I asked @RudiThoeni if I could do some merges during the weekend, he said ok, so if you want I can merge this PR

gappc:
One thing came to my mind: would it be possible to extract the `delete` implementation from [TableView.vue](https://github.com/noi-techpark/it.bz.opendatahub.databrowser/blob/0ada02e1b581fbe858c3a795deb6ccae5dbaffb7/databrowser/src/domain/datasets/ui/tableView/TableView.vue) into something like `useTableDelete`? That would move complexity out of the TableView.vue

a-crea:
@gappc would it cause problems to merge it as it is and then to apply the improvement? Each time I try to rebase it breaks the commits history... otherwise I could need your help to understand what is happening :/

I'm on Node 18 as suggested

gappc:
@a-crea I think we can merge it now and apply the improvements afterwards. @RudiThoeni could you please merge this PR?

@a-crea given that rebasing causes trouble, we could schedule a meeting where we can take a look at how to fix errors associated with it.

RudiThoeni:
done

gappc:
@a-crea could you please take a look at the problem? Thx

PR #492, opened 2023-12-15
Fix computation of deprecation information

gappc:


gappc:
Wrong commit message and PR title. I'll open up a new PR

Issue #423, opened 2023-09-19
As Open Data Hub Data Browser user I want that the deprecated fields of a dataset are hidden per default, to avoid using and modifying deprecated fileds.


sseppi:
As mentioned this morning in the meeting, thanks to a request processed by the Open Data Hub Customer Care team, it emerged the need to hide per default all fields marked as deprecated in the JSON Schema. The data browser user who wants to see also deprecated fileds has to activate it (maybe through a toggle). In that case the deprecated fields has to be clearly marked as deprecated and in addition the comment provided in the JSON Schema should be shown near the fields.

In the screenshot below you can see the example of the 3 "active" fields of the EventShord endpoint that has the label ```deprecated: true``` and in the red box there is the comment that should be shown if the deprecated fiels becomes active.

file:///home/stefano/Pictures/Screenshots/Screenshot%20from%202023-09-18%2016-24-11.png

In the screenshot below a I tried to show what should be hidden (fileds in the red box) and I added the toggle to explain better what I have in mind.

![Image](https://github.com/noi-techpark/it.bz.opendatahub.databrowser/assets/7909989/5277a276-aa21-4fa8-802e-784d5f376ec8)

**Note**: It is important that, if the ```Show deprecated``` is true, all deprecated fields are clearla shown as deprecated including the comment provided in the JSON Schema.

pkritzinger:
UI-Design for deprecated attributes:

https://www.figma.com/file/DmuP6Dbv5LzkCNOXrDnWIL/2023?type=design&node-id=1383%3A2538&mode=design&t=CW1pfW1uwovZEbMn-1

pkritzinger:
@sseppi @gappc we updated the design:
https://www.figma.com/file/DmuP6Dbv5LzkCNOXrDnWIL/2023?type=design&node-id=1383%3A2538&mode=design&t=CW1pfW1uwovZEbMn-1

Regarding color: we prefer the viola version, because it looks a bit fresher - but it's up to you to decide.

a-crea:
Hi @sseppi ,
I have a doubt about the implementation: where should I check whether a field is deprecated? Regarding your example, https://api.tourism.testingmachine.eu/v1/EventShort/eventshort-11378, I cannot find where to get a field marked as deprecated.

Can you provide me more info?

sseppi:
Hi @a-crea ,

[...]

PR #498, opened 2023-12-18
feat: add style for deprecated fields

a-crea:
This PR refers to #423

### DIFF
--- a/databrowser/src/components/tag/TagCustom.vue
+++ b/databrowser/src/components/tag/TagCustom.vue
@@ -74,6 +74,11 @@ const types: Record<TagType, Color> = {
     text: 'text-black',
     dot: 'bg-black',
   },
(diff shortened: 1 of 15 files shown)


gappc:
@a-crea looks really nice, good work :+1: 

One feedback: on mobile, in the detail / edit view, the toggles for `empy` and `depreacted` are not visible. Not sure if that's a bug or intentional.

Desktop (visible):
image


Mobile (not visible):
image

RudiThoeni:
is this PR ready to merge?

a-crea:
@gappc thanks :D It was intentional, show empty field was hidden on mobile, so I coded show deprecated the same way :) 
I would merge it as it is for now, we can discuss how to implement a mobile version whenever is needed.

PR #500, opened 2023-12-18
Use refetch from table data load

gappc:
@a-crea this PR builds upon your PR #499 and adds a single change: the `refetch` function now comes from the table data loader (in effect, it comes from the underlying useQuery). That way we don't need to set the URL query param  - nice idea btw ;)

The important changes can be found here:

https://github.com/noi-techpark/it.bz.opendatahub.databrowser/commit/1bdb4fb128224a855f04d05c0349027802ef1f31#diff-dd92aee5aa9e0482de571eb6a2370cbb07a5a21876eaa328b7f1956a97bb1a29R18

and

https://github.com/noi-techpark/it.bz.opendatahub.databrowser/commit/1bdb4fb128224a855f04d05c0349027802ef1f31#diff-6c6021849e6fff84340b416a6146e8d9000555c2a2ca55a28ae5a05b8ba5dee6R29-R52

The rest of PR #499 looks good to me.

If the changes are ok for you, we could take the shortcut and just merge this PR


RudiThoeni:
is this PR ready to merge?

a-crea:
@gappc I tested it again, it works fine and the code is cleaner, thank you! :)

Issue #501, opened 2023-12-21
Behaviour of filters in Table view


pkritzinger:

pkritzinger:
If I define multiple filters in the table view and click the button "Filter", only the respective filter is activated. I would expect that all defined filters are activated, regardless which of the buttons is clicked.

From my point of view we have two options:
- Activate all filters on click, regardless of button clicked
- Integrate filter --> Apply all filters


@sseppi could you pls. plan this task accordingly, so that we discuss it (we should fix it before the UX-Test)



![Image](https://github.com/noi-techpark/it.bz.opendatahub.databrowser/assets/92454516/901acfbc-c307-4a4d-abd3-4a9868a932ae)

gappc:
@pkritzinger that seems to be a bug, could you please provide a link?

pkritzinger:
@gappc https://databrowser.opendatahub.testingmachine.eu/dataset/table/tourism/v1/EventShort?rawfilter=and(eq(EventLocation,%27NOI%27),ge(StartDate,%2705.12.2017%27),le(EndDate,%2722.01.2018%27))

this is the link with all filters set (I clicked all 3 buttons to set them)

gappc:
@sseppi @pkritzinger should be fixed with #506, please check

sseppi:
@pkritzinger now the filter button seems to work as defined by you. Do you confirm?

pkritzinger:
@sseppi yes, works from my POV

PR #506, opened 2024-01-12
Fix table filter

gappc:
Fix and improve table filter (see issue #501 for bug description)


gappc:
@RudiThoeni @dulvui from my point of view this PR is ready to be merged

PR #511, opened 2024-01-13
feat: gps point editable only in edit view

a-crea:
This PR sets gps fields as not editable in GPS tab while in detail view.

Issue #515, opened 2024-01-14
GPS map: data for GPS point is not reset when changes are canceled


gappc:
When editing GPS point data, changes are not reset when clicking on the "Cancel" button, although the data seems to be reset correctly (see screenshot below).

This problem effects the inputs as well as the location on the map.

Steps to reproduce:

- open https://databrowser.opendatahub.testingmachine.eu/dataset/detail/tourism/v1/Accommodation/5CEA544EE34639034F07B79D4AEEB603#gps-data
- click on the edit symbol to edit a GPS Point
- change e.g. the `latitude` => the "Save" and "Cancel" buttons on at the bottom should now be visible
- click on "Cancel"

=> it is expected, that the changes are canceled and the original data is restored. Therefor, it is expected, that the latitude is restored and the position icon on the map is moved, too. This does not happen, although the data seems to be reset correctly (take a look at the JSON DIFF tool in the toolbox on the right).

Note that the underlying implementation (`EditListCell.vue` and related components) seems to work fine, because trying to reset image data works as expected (e.g. here: https://databrowser.opendatahub.testingmachine.eu/dataset/edit/tourism/v1/Accommodation/5CEA544EE34639034F07B79D4AEEB603#images)

![image](https://github.com/noi-techpark/it.bz.opendatahub.databrowser/assets/5732768/e97e6a35-a083-4ce2-8eeb-10431b784161)

Issue #514, opened 2024-01-14
GPS map: duplication of entry does not work


gappc:
Duplication of GPS map entries does not work and throws an error (see screenshot below)

Steps to reproduce:

- open https://databrowser.opendatahub.testingmachine.eu/dataset/edit/tourism/v1/Accommodation/5CEA544EE34639034F07B79D4AEEB603#gps-data
- click on the three-dots icon right to the GPS entry => a menu is shown
- click on "Duplicate"

=> it is expected, that the entry is duplicated. Instead, nothing happens and an error is shown in the console

Note that the underlying implementation (`EditListCell.vue` and related components) seems to work fine, because trying to duplicate e.g. images works as expected (e.g. here: https://databrowser.opendatahub.testingmachine.eu/dataset/edit/tourism/v1/Accommodation/5CEA544EE34639034F07B79D4AEEB603#images)

![image](https://github.com/noi-techpark/it.bz.opendatahub.databrowser/assets/5732768/8628fc0c-04c8-4fcf-8afc-b62e050945e0)

PR #519, opened 2024-01-19
fix: reactive editable data for gps points. Duplicate and add new ite…

a-crea:
This pull request addresses issues #514 and #515.
Additionally, it supersedes pull request #511, as the features introduced in #511 are now incorporated in this pull request.


RudiThoeni:
ok then i close #511 and merge this one?

Issue #536, opened 2024-02-26
Add new room in Event short doesn't work


d3liaa:
Browser: Chrome
**Problem Description:**
In the dataset EventShort - > add a record  ->  event detail (https://databrowser.opendatahub.com/dataset/new/tourism/v1/EventShort#event-details) 
When i click on add new room i expect to appear a table like in Technology Fields and Tagging Fields but nothing happens. But the bar with cancel or save does appear.(inserted screenshot below)
I also tried to add a new room to an already existing event, and in this case nothing happened , not even the cancel or save bar.
(ex: https://databrowser.opendatahub.com/dataset/edit/tourism/v1/EventShort/eventshort-11378#event-details)

Screenshot:
![Screenshot 2024-02-26 165711](https://github.com/noi-techpark/it.bz.opendatahub.databrowser/assets/101118017/361ddd69-7eb6-42bf-bc5d-34fd67f9ba64)

sseppi:
Hi @gappc,

I have the same problem too.

I post below also my console screenshot, where I see 2 warnings. But I think they aren't related to this problem.

![Screenshot from 2024-02-26 17-31-19](https://github.com/noi-techpark/it.bz.opendatahub.databrowser/assets/7909989/86f318aa-591e-4c5b-bc50-3ef20fd0e295)

Can you please check?

gappc:
@d3liaa @sseppi PR #542 should fix the issue, please take a look after the merge is done and close this issue if the bug was resolved

sseppi:
I tested the functionality and it seems to work fine.

@RudiThoeni did you already released the PR #542, if not can you please release it on production too?

RudiThoeni:
done

sseppi:
@RudiThoeni thank you!

PR #542, opened 2024-03-05
Generify edit list cells

gappc:
Generify components based on edit-list-cell, together with the action hooks and triggers

This commit also fixes issue #536


gappc:
@RudiThoeni ready to merge

PR #546, opened 2024-03-13
PR542

RudiThoeni:


gappc:
@RudiThoeni please don't forget to merge this merge commit back to the development branch, otherwise we have different commit histories between prod and development branches, which in my opinion could be problematic in the future (e.g. when we want to compare histories / commits).

The goal is to have the same commit history in the prod and development branch.

Of course, this is also a matter of taste, so please let me know if you have another opinion on this.

Issue #534, opened 2024-02-13
As a Content Editor I want a "Push" button, to send push notification about specific contents I'm managing.


sseppi:
In the old Data Browser there was the possibility to send push notification about specific contents.

![Image](https://github.com/noi-techpark/it.bz.opendatahub.databrowser/assets/7909989/11419694-d442-4542-ad4b-af5fe2378167)

The first real use case we have is the Push of notification about NOI News to the NOI App.

In this issue we should define and implement a generic solution, to allows us to activate this push functionality without the need to riprogramm it every time there is a new app or content that should be pushd or notified.

**Open Questions**

* How do we define the user role that is allowed to send the push notification?
   Maybe it could be simply all users who have edit rights on the data);
* How do we set the apps that should be notified? Do we use the ```published on``` attribute?

sseppi:
@mrabans @RudiThoeni I've created this issue and we need to start defining how should be the generic solution. Are there other open questions that should be taken into consideration for this issue?

mrabans:
@sseppi probably we have to discuss this in detail to come up with a future proof solution.

RudiThoeni:
We discussed about the push functionality and the idea was to have one generic Push Endpoint where this informations has to be passed:

- id
- type of the dataset (example for tourism is _Meta.Type)
- language (if the push has to be sent only in one language)
- identifier --> could be the publisher example noi-communityapp

On Publisher Level the Push Functionality has to be activated. We thought about activating it on Dataset type.

The information (Timestap when, PushId etc...) has to ba saved on data level and displayed to the user somewhere, so he knows that a push was already sent etc....

RudiThoeni:
We had a discussion where we defined:

- We save the information (Push URL, Parameter etc...) into the publisher data
- Parameter are saved single (in style of the Metdata api)
- Parameters which should be filled by Databrowser are marked with `{fieldname}` example `{_Meta.Type}`
- Supported Parameters are simple datatypes, only string arrays are supported where all values are passed with a , as separator
- Where the possibility exist to process it on backend side (like the language on the noi community app push) we do it on backendside
- The Pushresult is saved in the data (date, result Push api,  etc..)
- We use HTTP POST instead of HTTP Get 

Some UX questions emerged

- Where should the button be displayed (Listview, Editview ...)
- What if more Push buttons are there (Button with Dropdown)

pkritzinger:
@RudiThoeni what do you mean with 

[...]

PR #556, opened 2024-04-07
Implement push notification

gappc:
@sseppi @mrabans @RudiThoeni @pkritzinger @Mazzolintis: This PR implements the push notification component as defined in  #534 

You can find a demo version here: https://9.databrowser.gappc.net/dataset/table/tourism/v1/ODHActivityPoi

> Note that the column for the push functionality is the last one in the table

image

In order to take a look at the push responses, I've also added a config for them. A demo for the push responses can be found here: https://9.databrowser.gappc.net/dataset/table/tourism/v1/PushResponse

Please take a look at the PR and tell me your opinion


gappc:
Open topics to discuss:

- the push button in the table view is disabled by default. At the moment, it is enabled only if the user is logged in. Should we restrict enabling the button further e.g. using roles?
- after a push is send, the "Send out the Push-Notifications" button is disabled. One needs to close the push-popup and then reopen it again to send another push. is this ok or should we add a "reset" button? If we don't want to have a "reset" button, should we add a description / text that explains how a user could do another push?

RudiThoeni:
Hi @gappc 
Grea work!
For the points to discuss on my view 
-I think enabling the button for the logged user should be enough..... the api then gives 401/403 etc... if the user is not allowed. Lets try this easy solution and if we see that we have to add role based button hiding we cann add it later...
-For me this was clear, to close the popup and reopen it, for me as it works good...... but i let decide here stefano + the ux experts

On thing that came in my mind 
-Instead of showing the OPEN DATA HUB STATE in the table view i will change the config to show the "PublishedOn" published Channels, everyone agrees?

sseppi:
Hi @gappc and @RudiThoeni 

great work! I like it.
> Hi @gappc Grea work! For the points to discuss on my view -I think enabling the button for the logged user should be enough..... the api then gives 401/403 etc... if the user is not allowed. Lets try this easy solution and if we see that we have to add role based button hiding we cann add it later...

I agree, let keep it simple. In case the real users get confused, we will then update. It is important that we expose a clear error description.

 -For me this was clear, to close the popup and reopen it, for me as it works good...... but i let decide here stefano + the ux experts

Also here I agree with Rudi.

> 
> On thing that came in my mind -Instead of showing the OPEN DATA HUB STATE in the table view i will change the config to show the "PublishedOn" published Channels, everyone agrees?

Fine for me!

pkritzinger:
@gappc awesome, thanks!

>after a push is send, the "Send out the Push-Notifications" button is disabled. One needs to close the push-popup and then reopen it again to send another push. is this ok or should we add a "reset" button? If we don't want to have a "reset" button, should we add a description / text that explains how a user could do another push?

I also think that it's pretty straightforward, one optimization could be that we do not show the button in the disabled state anymore (user may think that (s)he needs to do sth. in order to reactivate the button) but only show the success msg?

what do @sseppi and @RudiThoeni  think?

sseppi:
@gappc @pkritzinger @RudiThoeni 

As mentioned during the meeting, I think we can move this functionality in testing.

I have only small comments I would share with you:

1. I suggest to change the text in the button from "Send Push Notification" to "Send Push"
2. Once the Push has been sent, I would suggest to add in the cell of the Table View and the Popup the ID and the timestamp of the sent Push
3. In case of error, we need to notify the user if the error is due to technical problems or missing rights.

I would suggest to close this PR and for my comments open a dedicated issue.

What do you think?

gappc:
@sseppi yes, I agree, let me (or @RudiThoeni) know when the PR should be merged

RudiThoeni:
then lets merge it?

sseppi:
> then lets merge it?

From my point of view yes, so I can show to our communication

RudiThoeni (approved):

PR #567, opened 2024-04-23
Allow all publishers

gappc:
The primary task of this PR is to enable the noi-community-app publisher (was removed from list of publishers)

Other improvements:
- show info if no publishers are available
- improve popover component handling


RudiThoeni:
@gappc sorry i replied your email before seeing this PR will now merge and test it

PR #592, opened 2024-08-15
Fix cancel / save button position for mobile view

gappc:
The cancel / save buttons in the EditView were visible only at the bottom of the mobile view.

Depending on the height of the page, they were only visible after a user scrolled down.

This commit fixes that issue.

### DIFF
--- a/databrowser/src/domain/datasets/ui/editView/EditView.vue
+++ b/databrowser/src/domain/datasets/ui/editView/EditView.vue
@@ -54,7 +54,7 @@ SPDX-License-Identifier: AGPL-3.0-or-later
           <EditToolBox />
         </div>
         <EditFooter
-          class="transition-all"
+          class="sticky bottom-0 transition-all"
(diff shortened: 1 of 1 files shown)


gappc:
@RudiThoeni ready for merge

RudiThoeni:
merged

sseppi:
@gappc I tried in testing but I still miss the two buttons if I edit something. Is it already in testing the feature or should I use another staging link?

gappc:
@sseppi strange, I just tested it on my mobile phone and the buttons are shown. Are they maybe they hidden behind the keyboard?

Tested in the usual development stage, e.g. https://databrowser.opendatahub.testingmachine.eu/dataset/edit/tourism/v1/Accommodation/5E4BEFD440200CE4732A7FA30897C402#contact

gappc:
@sseppi update: I can now confirm the behaviour you described, I need to take another look at it, sry.

PR #612, opened 2024-10-03
Fix/optimizations and fixes pr 596

a-crea:
This PR fixes refer to https://github.com/noi-techpark/it.bz.opendatahub.databrowser/pull/596#issuecomment-2381437866

Remaining points for discussion:
- Referenced attributes not displaying when the toggle is disabled
- Potential confusion with the toggle text: the wording has been changed, but needs confirmation
- Deciding between using a Popover or Tooltip component"

### DIFF
--- a/databrowser/guideline.md
+++ b/databrowser/guideline.md
@@ -14,6 +14,7 @@ If something is not specified below, it is advisable to try to search for a simi
 
 - [Coding Style](#coding-style)
 - [Platform assumptions](#platform-assumptions)
+- [Deprecations](#deprecations)
 - [Reuse Compliance](#reuse-compliance)
 
 ### Coding Style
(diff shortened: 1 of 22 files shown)


a-crea:
@gappc, in guideline.md, should I mention avoiding TooltipCustom because it will be deprecated (and use Popover instead) under the "Coding Style" section, or should I create a new section, such as "Additional Information"?

gappc:
@a-crea good question. What you say if we create a section "deprecations" (or something like that) and add there for any  deprecation information a subsection where we shortly explain the what, why and alternative?

a-crea:
@gappc guidelines updated :) 
If you agree I think we can consider the PR ready to merge :)

sseppi:
@gappc I think that can be merged too, please notify @RudiThoeni if you agree too,

Issue #574, opened 2024-06-21
Traffic Station dataset doesn't work properly


sseppi:
I tried to access the Traffic Station dataset (see the following link) but I experinsed some problems.

https://databrowser.opendatahub.com/dataset/table/mobility/v2/flat/Trafficstation

It seems that the Data Browser doesn't shows all traffic stations in the Table View.

- **Device**: Desktop
- **Operating System:** Ubuntu 24.04
- **Browser:** Chrome (v 125.0.6422.141), Brave (1.67.119) and Firefox (127.0.1)
- **Problem Description:**
Accessing the data about the Traffic Station only the first 25 Station are loaded and the "Lines per page" dropdown shows a strange "Infinity" message. Ad the switcher to next page doesn't work

![Screenshot from 2024-06-21 09-14-54](https://github.com/noi-techpark/it.bz.opendatahub.databrowser/assets/7909989/b3f9dfa7-930a-4e1c-9518-a80a4e33d9cc)

- **Expected behaviour:**
I would like that the user experience of the switch to next page and the setup of the lines per page, is the same for all dataset (see screenshot below). In case that the Dataset has few entry the user should have anyway the possibility to choose different page sizes and in case be notified that there aren't more pages to navigate.

 ![Screenshot from 2024-06-21 09-20-49](https://github.com/noi-techpark/it.bz.opendatahub.databrowser/assets/7909989/75caccd7-5cfb-4c42-afec-08a64ee49a6e)

gappc:
@sseppi the Traffic Station dataset (like all mobility datasets) uses a pagination type that is different than the one from the tourism domain. The response does not contain any information about how many records are available at all, so there is no way to show that information to the user.

For the Traffic Station dataset pagination we pass the parameters `offset` and `limit` with the request, as a result we get an object with the following properties:

- `data`: contains an array of result records (25 by default)
- `offset`: the offset used for the request
- `limit`: the limit used for the requests

Regarding the other feedback "...the user should have anyway the possibility to choose different page sizes...": this is already possible using the "Lines per page" option. Did you mean something different?

sseppi:
@gappc thank you for your feedback!

gappc:
@sseppi sorry, at first I didn't understand what you mean. Now I've fixed the problem with PR #615

[...]

PR #615, opened 2024-10-03
Fix mobility pagination bug

gappc:
Fix issue #574


gappc:
@sseppi if the fix is ok (see also [https://2.databrowser.gappc.net/dataset/table/mobility/v2/flat/Trafficstation])(https://2.databrowser.gappc.net/dataset/table/mobility/v2/flat/Trafficstation), then we can merge this PR.

Issue #663, opened 2024-12-12
BUG: link in tooltip doesn't work


sseppi:
If I click on a button included in a tool-tip doesn't work, since there isn't any reaction if I click on it.

**To reproduce the problem:**

1. open a table view in the testing data browser: https://databrowser.opendatahub.testingmachine.eu/dataset/table/tourism/v1/Accommodation
2. open the filter side bar
3. click on the info icon near "ALL FILTERS"
![Image](https://github.com/user-attachments/assets/9d6b546c-e580-4f4f-bb10-d637d5baebe1)
4. click on the "Open wiki" button
![Image](https://github.com/user-attachments/assets/6a809933-78f0-4426-a2d6-d9e693174501)

**Expected behaviour**
The right wiki page must be opened in a new window.

gappc:
@sseppi PR #668 fixes this issue

RudiThoeni:
works can be closed

sseppi:
@Xeroxi91 the bug has been solved, now everything should work fine in the testing environment.

https://databrowser.opendatahub.testingmachine.eu/

PR #668, opened 2024-12-19
Bugfix: link in tooltip

gappc:
This PR fixes issue #663


RudiThoeni:
@gappc @sseppi  i tested it works we can close the issue should i deploy to prod?

gappc:
@RudiThoeni in my opinion it can be deployed to prod

PR #671, opened 2025-01-08
Switch to maplibre as library for maps

gappc:
This commit removes "leaflet" and introduces "maplibre" as library for maps. The reason is, that the next tasks for the Data Browser requires a more capable map library.


RudiThoeni:
@gappc deployed to test, i tested and for me the map seems working fine (tried on GPS Points Detail/EditView), please have a look and tell me if i can proceed to move to production

PR #693, opened 2025-02-14
Fix dataset URL computation for map

gappc:
The old computation did not account for API urls that already include query params.


gappc:
@RudiThoeni ready for merge

Issue #650, opened 2024-10-24
As Data Browser user I want a better management on the context, to come back to the same view where I was once I click on the back button


sseppi:

sseppi:
It seems that the context is working much better now.
I close the issue and we will open a dedicated issue if we find some edge case to improve.

PR #696, opened 2025-02-19
Resolve Issue #650

mariapaolasmnt:


gappc:
@mariapaolasmnt thank you for reporting the problem, I've put it in a new issue #699 for discussion.

RudiThoeni:
please add the spdx header on each new file ;)

RudiThoeni:
@gappc  is this PR ready to merge or wait for the discussion you mentioned?

mariapaolasmnt:
> @gappc is this PR ready to merge or wait for the discussion you mentioned?

We discussed it with @gappc and are currently working on it. We'll let you know once it's ready

sseppi:
@gappc can we merge this PR?

databrowser/src/pages/datasets/overview/OverviewListPage.vue
```diff
@@ -366,12 +383,50 @@ const hideFilters = () => {
   isFiltersModalVisible.value = false;
 };
 
+const filterSelectedForComponent = ref<{ key: string; value: string }[]>([]);
+
+const initializeFiltersAndSearch = () => {
+  const {rawfilter, searchQuery} = getStartedQuery();
+
+  if (!searchQuery && !rawfilter) {
```
gappc, line 391: Watch out for implicit type casts, are you sure that any developer that touches this code knows whats happening? I bet that's not the case ;)

So in my opinion it's better to do explicit comparison, e.g. `searchQuery.length === 0 && rawfilter.length === 0` or similar

databrowser/src/pages/datasets/overview/OverviewListPage.vue
```diff
@@ -366,12 +383,50 @@ const hideFilters = () => {
...
+    return;
+  }
+
+  if (searchQuery && searchQuery !== '') {
```
gappc, line 395: Isn't this just a check for string emptiness? Watch out for implicit type casts...

databrowser/src/pages/datasets/overview/OverviewListPage.vue
```diff
@@ -366,12 +383,50 @@ const hideFilters = () => {
...
+    filters.value.searchVal = (searchQuery as string) || '';
```
gappc, line 396: `searchQuery` is already known to be a string, no need to cast it to `string` again.

In addition, here I'm not 100% sure what you try to express with `... || ''`, `searchQuery` is typed as `string`, so the `||` will never be evaluated.

If `searchQuery` can be `undefined`, `null` or something else, then the types are not correct and you need to fix them.

databrowser/src/pages/datasets/overview/OverviewListPage.vue
```diff
@@ -366,12 +383,50 @@ const hideFilters = () => {
...
+    if (!rawfilter) updateURL(router, [], searchQuery);
```
gappc, line 397: Watch out for these type casts, please do explicit comparisons

databrowser/src/pages/datasets/overview/OverviewListPage.vue
```diff
@@ -366,12 +383,50 @@ const hideFilters = () => {
...
+  }
+  if (rawfilter) {
```
gappc, line 399: Please do explicit comparisons

databrowser/src/domain/homepage/utils.ts
```diff
@@ -0,0 +1,40 @@
...
+//
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+import {Router} from "vue-router";
+import {useQueryStore} from "./filterStore.ts";
+import {useRouteQuery} from "@vueuse/router";
+
+const queryStore = useQueryStore();
+
+export function updateURL(router: Router, filters: string[] = [], searchQuery: string = '') {
```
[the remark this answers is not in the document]
mariapaolasmnt: i provide router as param because this is a utils, i already tried with the  useRouter() but it didn't work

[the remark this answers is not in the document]
gappc: Hmmm that's strange, I use that pattern in several places, e.g. https://github.com/offline-agency/it.bz.opendatahub.databrowser/blob/feature/databrowser/src/domain/datasets/ui/mapView/detail/useViewElements.ts#L19 or https://github.com/offline-agency/it.bz.opendatahub.databrowser/blob/feature/databrowser/src/domain/datasets/ui/tableView/useTableRowSelection.ts#L13

Could you please give it another try and let me know? Passing the router is ok if there is no other way to do it, otherwise I'd rather not do it, because Vue3 has the concept of composables that fence such stuff.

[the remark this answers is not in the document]
gappc: @mariapaolasimionato I just tested it and it does indeed not work. I'll take a look at it and let you know, you can ignore this feedback as of now.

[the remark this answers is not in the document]
gappc: @mariapaolasimionato ok, got it. The reason it does not work is, that the `updateURL` invocation is performed in the `toggleFilter` function of the `OverviewListPage` component. That does indeed not work as expected, you'll get the same error if you do e.g. the following:

```typescript
const toggleFilter = (key: string, value?: string) => {
  // ... other code
  const router = useRouter();
  updateURL(router, updatedFilters.value, filters.value.searchVal);
}
```

The thing is, that `useRouter` can not be used inside event callbacks and e.g. `computed(() => ...)` functions. You would need to adjust the implementation, putting the `updateURL` function into the outermost place of `OverviewListPage` and then working with `ref` values. I think that goes to far here, so just leave this as you implemented it - but please still take into account the other comments from the code review.

I'm going to mark this conversation as resolved.

databrowser/src/pages/datasets/overview/OverviewListPage.vue
```diff
@@ -218,14 +234,13 @@ SPDX-License-Identifier: AGPL-3.0-or-later
 </template>
 
 <script setup lang="ts">
-import { computed, ref } from 'vue';
+import { computed, onBeforeMount, ref } from 'vue';
```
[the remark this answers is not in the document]
gappc: Ok, then leave it there :+1:

mariapaolasmnt:
@gappc you can check the PR

databrowser/src/components/toggle/ToggleCustomHomePage.vue
gappc, line 1: @mariapaolasimionato - code formatting ;)

databrowser/src/domain/homepage/utils.ts
```diff
@@ -0,0 +1,47 @@
...
+import { useRouteQuery } from '@vueuse/router';
+
+import * as R from 'ramda';
+
+export function useUpdateURL(router: Router, filters: string[] = [], searchQuery: string = '') {
+  const currentQuery = router.currentRoute.value.query;
+  const newQuery: { [key: string]: string | LocationQueryValue[] | null } = {
+    ...currentQuery,
+  };
+  if (!filters || filters.length > 0) {
```
gappc, line 15: What does the `!filter` check? `filter` is typed as array, so a length check should suffice

databrowser/src/domain/homepage/utils.ts
```diff
@@ -0,0 +1,47 @@
...
+    newQuery['search'] = searchQuery;
+  } else {
+    delete newQuery['search'];
+  }
+
+  if (!R.equals(newQuery, currentQuery)) {
+    return router.push({ query: newQuery });
+  }
+
+  return;
```
gappc, line 31: No need to `return` here

gappc (commented):
@mariapaolasimionato some minor stuff, then this PR should be ready

gappc:
@mariapaolasmnt this PR has merge conflicts, please resolve them, then we should be able to merge this PR.

RudiThoeni:
merged

PR #716, opened 2025-03-20
Integrating UpdateHistory

RudiThoeni:
This PR integrates UpdateHistory Detailview and some changes on EditListTable, 
By adding the parameters `hideSelectCheckbox` and `hideTableActionHeader` the table view can be used in readonly mode


RudiThoeni:
@gappc please have a look at this, i hope it is ok if i add some properties to the EditListTable because i want to use it in readonly mode also in the edit view....

gappc:
@RudiThoeni there are small issues, but the PR looks quite good to me. Here is some feedback:

- please format the code according to prettier / linter
- although I see the advantages of your implementation (fine-grained control of `EditListTable` UI), the changes to `EditListTable` are not strictly necessary. You get the same effect with the piece of code shown below. In this particular case, it should be added to the `UpdateHistoryTable` component

```typescript
// Set edit mode to false for this and all child components
useProvideEditMode(ref(false));
```

Not sure about `EditListTable`, what is your opinion?

gappc:
@RudiThoeni that's fine with me :+1:

RudiThoeni:
@gappc for the code formatting issues i always use `npm run lint -- --fix` / `npm run lint:fix` before committing..... code should be already formatted? or am i missing some linting command?

gappc:
@RudiThoeni the command you run invokes just the linter [eslint,](https://eslint.org/) but does not run the code formatter [prettier](https://prettier.io/). See here for a simple comparison between linter and code formatter: [https://prettier.io/docs/comparison/](https://prettier.io/docs/comparison/).

Usually, you let your IDE run the code formatter in an automatic way, e.g. on save. Here is the list of IDEs supported by prettier: [https://prettier.io/docs/editors](https://prettier.io/docs/editors).

If you want to run prettier on the command line, then you can find a list of commands [here](https://prettier.io/docs/cli), e.g. `prettier . --check` to check for files that don't adhere to the code formatting rules.

Note that `prettier` is the default tool for code formatting on the frontend (since some years), but given that more code formatting issues pop up recently, should we add a `package.json` script and git hook to streamline code formatting?

RudiThoeni:
> @RudiThoeni the command you run invokes just the linter [eslint,](https://eslint.org/) but does not run the code formatter [prettier](https://prettier.io/). See here for a simple comparison between linter and code formatter: https://prettier.io/docs/comparison/.
> 
> Usually, you let your IDE run the code formatter in an automatic way, e.g. on save. Here is the list of IDEs supported by prettier: https://prettier.io/docs/editors.
> 
> If you want to run prettier on the command line, then you can find a list of commands [here](https://prettier.io/docs/cli), e.g. `prettier . --check` to check for files that don't adhere to the code formatting rules.
> 
> Note that `prettier` is the default tool for code formatting on the frontend (since some years), but given that more code formatting issues pop up recently, should we add a `package.json` script and git hook to streamline code formatting?

thx for your reply, maybe a hook to format the code would be great?
When i ran `prettier . --check` a lot of files where listed.... also files which i didn't touched.....

Issue #695, opened 2025-02-18
Quick Search in Detail Mode does not display right dataset


RudiThoeni:
If i click from tableview into a record

![Image](https://github.com/user-attachments/assets/1fe72c0a-245f-445a-8498-b5085044e9ab)

the quicksearch changes

![Image](https://github.com/user-attachments/assets/d588f1af-b651-477c-affa-acb41da34624)

RudiThoeni:
ok i saw this is only the case of my example;)
The reason is that the Metadata of Noi News has the apiurl "https://tourism.api.opendatahub.com/v1/Article?articletype=newsfeednoi"
and there is some search finding ArticleType as Entity........ what could we do to avoid this?

sseppi:
@RudiThoeni thank you for finding this out and sharing it.

I qill check with @gappc and try to find a solution

gappc:
@RudiThoeni I could reproduce the issue on the current PROD environment, opening https://databrowser.opendatahub.com/dataset/table/tourism/v1/Article?articletype=newsfeednoi and then clicking on `DETAIL`, but I was not able to reproduce it in the current TEST environment, opening https://databrowser.opendatahub.testingmachine.eu/dataset/table/tourism/v1/Article?articletype=newsfeednoi

It seems that the bug got resolved in PR #602 or some time afterwards, but I'm not really happy with the solution, because it relies on session storage. You therefor need follow certain steps to get to reproducible results, which leads to subtle bugs.

Examples:

### Variant 1 (works as expected)

- open up the table view https://databrowser.opendatahub.testingmachine.eu/dataset/table/tourism/v1/Article?articletype=newsfeednoi => the dataset name is "NOI news" as expected. Note the `articletype=newsfeednoi` query param. In this table view here, the session storage gets set (see image below)

![Image](https://github.com/user-attachments/assets/27443aab-c9b3-4556-8f1e-a02a9dbf90e5)

- click on `DETAIL` of any record (e.g. 
https://databrowser.opendatahub.testingmachine.eu/dataset/detail/tourism/v1/Article/4A418EFC-9EBA-0D28-C0CD-EA289006239C) => the dataset name remains "NOI news" as expected. Note that the `articletype=newsfeednoi` query param is missing in the URL, this required information is taken from the session storage that had to be set on the table view before. If it would not be set, the dataset name would be "Articles"
 
### Variant 2 (buggy)

[...]

PR #718, opened 2025-03-21
Bugfix/handle subdataset in url

gappc:
This PR fixes issue #695 and a minor bug, where pagination information got lost when the user navigated from table- to detail-, and then back to table-view.


gappc:
@RudiThoeni this PR is ready to be merged, thx

Issue #689, opened 2025-02-06
As a Quality manager I want to open clustered lists in the table view


sseppi:
The map generates clusters when many data points are visualized and the user zooms out. Clicking on a cluster opens the info box with the list of records and their short name.

For quality tasks it would be super useful to open this list in the table view to be able to work on it, filter it and perhaps export it.

sseppi:
We have to consider that, in the near future we will have also an Open Data Hub data Quality tool. and we need to elaborate also the connection with the other Open Data Hub tools.

gappc:
@sseppi you commented:

> We have to consider that, in the near future we will have also an Open Data Hub data Quality tool. and we need to elaborate also the connection with the other Open Data Hub tools.

Is this a general remark that applies also to this issue, or is it a requirement just for this issue? In the former case, I'd create a new issue, because otherwise I think the requirement will get buried in this thread. Generally, connecting different tools involves having linkable resources (URLs), which will take us a long way.

About the topic "clustered lists in the table view": I have to take a look at the API's, I think there is a chance that we get the cluster bounding boxes from the map and reuse that for the Open Data Hub requests, but I need to do some testing., Should I take a look at it?

gappc:
@sseppi i did now some experimentation with the content API, here are my findings.

First, it's possible to get the cluster bounding boxes and use that information to show the records of a cluster in the table view.

For example, the following link opens the map and selects the "Accommodation Mountain" dataset: https://databrowser.opendatahub.testingmachine.eu/?map=true&datasetIds=93317701-2ffa-4a6c-9fee-5d15a28a93ed

This gives 3 clusters with 19, 31 and 50 records (see image below)

![Image](https://github.com/user-attachments/assets/bfbd2e5e-6376-4ce0-8461-bb96017a0e59)

Here are the links with bounding boxes for the table view:
 
- 19 records: https://databrowser.opendatahub.testingmachine.eu/dataset/table/tourism/v1/Accommodation?typefilter=32&polygon=bbc(11.752088%2046.7017,11.752088%2046.364104,11.197098%2046.364104,11.197098%2046.7017,11.752088%2046.7017)
- 31 records: https://databrowser.opendatahub.testingmachine.eu/dataset/table/tourism/v1/Accommodation?typefilter=32&polygon=bbi(11.402639%2046.973781,11.402639%2046.500844,10.435467%2046.500844,10.435467%2046.973781,11.402639%2046.973781)
[...]

PR #723, opened 2025-03-27
Fix link from clustered records in map view to table view for timeseries API

gappc:
This PR fixes the link building from the map to the table view for `timeseries` style APIs (see also #689)


gappc:
@RudiThoeni this PR is ready to be merged

PR #761, opened 2025-05-17
Bugfix/map detail and linking for mobility

gappc:
This PR fixes two issues with the map view, both related to the timeseries API:

- fix the request URL for a single timeseries record
- fix timeseries response data handling


gappc:
@RudiThoeni this PR is ready for merge

PR #774, opened 2025-07-10
Update dependencies 2025-07

gappc:
This PR updates all project dependencies to their current versions as of 2025-07, with the following exceptions:

- **ramda**: not updated from 0.30.1 to 0.31.3 because is duplicates the modules that need to be compiled by vite and the bundle size  slightly increases. Since there is no pressing reason to update it, this package is left at its current version
- **tailwindcss**: not updated from 3.4.17 to 4.1.11 because a quick check showed, that more work needs to be done. Since the UI is in the process of being restyled, it is best to do that update during the restyling, @sseppi what do you say?


gappc:
@RudiThoeni this PR is ready for merge

RudiThoeni:
merged

sseppi:
> * **tailwindcss**: not updated from 3.4.17 to 4.1.11 because a quick check showed, that more work needs to be done. Since the UI is in the process of being restyled, it is best to do that update during the restyling, @sseppi what do you say?

@gappc OK, fine for me to do the update while we do the restiling. Please remember it during the kick-off with the developers that will implement the restyling.

Issue #769, opened 2025-06-11
GpsTrack View


RudiThoeni:
I added a Tableview with the linked GPSTrack. Example https://databrowser.opendatahub.testingmachine.eu/dataset/detail/tourism/v1/ODHActivityPoi/routes-hikingtrails.10213329?source=civis.geoserver#gps-track

Would be cool if we can add a functionality like on the GPS Point where the track is displayed on a map..

The tracks can be differ in their format. Normally there are three formats available

- geojson
- gpx
- kml

I would start with geojson support

gappc:
@RudiThoeni @sseppi I'm on vacation for the next 3 weeks, but I can take a look at it afterwards, if you want.

lukmay:
@sseppi yes sure! GEO data is currently my most favorite one, so I’m happy to work on this :)
My schedule is quite full for the next two weeks. However, I can start at the end of next week or early the following week if that fits the timeline of this.

lukmay:
Hi @sseppi and @RudiThoeni,

I have finished the POC for the new map component that displays tracks. You can see it in action in the video below. I adapted the style from the GPS points component.

Local example URL:  
http://localhost:3000/dataset/edit/tourism/v1/ODHActivityPoi/routes-hikingtrails.10214498?source=civis.geoserver#gps-track

https://github.com/user-attachments/assets/ce42c2da-26e8-4b19-9a8c-e27ee9bb3955

The main issue right now is a CORS error when retrieving data from our `http://api.tourism.testingmachine.eu/v1/GeoShape` endpoint, since `http://localhost:3000` is not allowed in its configuration. The other endpoints like `https://api.tourism.testingmachine.eu/v1/Accommodation` and the production GeoShape endpoint `https://tourism.api.opendatahub.com/v1/GeoShapes` already accept requests from localhost.
(For the demo, I am proxying the API request through Vite as a local file, but I think requesting it directly the same way as all other API calls would be better)

@RudiThoeni could you please update the server’s CORS settings? Once that’s done, I can adapt to it and push the implementation to a new branch for review.

Let me know what you think of the design or if you’d like any changes.

sseppi:
@lukmay nice extension! In seems to me, you did a great job. I'm really looking forward to test it in the testing environment.

Maybe we could consider to:

- use the Open Data Hub green to design the track, instead of the blue you are using now;
- put a begin and end marker, to let the user understand where is be start and the end of the track.

[...]

PR #775, opened 2025-07-15
add new track map component

lukmay:
This PR implements the new Track-Map component requested in https://github.com/noi-techpark/opendatahub-databrowser/issues/769

The current implementation consists of:
- New TrackMap component integrated into GpsTrackTable and GpsTrackTab.
- Supports GeoJSON, basic GPX, and basic KML via custom parsing, prioritizing paths and track data; if those aren’t available, it also handles waypoints.
- Visuals adapted to the Open Data Hub branding.
- No start or end markings, since the beginning and end of a file don’t always correspond to the actual start and end of a track.

image

image

image


RudiThoeni:
all merged ;)
@lukmay very good work! thx

PR #807, opened 2025-11-25
Chinese Box, Wkt, MapView, Object as Array

Luscha:
- Created EditObjectAsListCell.vue component to flatten object to array to facilitate showing and editing of dictionary using already existing table/tab components
- Created EditableWktMap.vue map to properly display and draw free shapes handling wkt input and output
- Created editGeoDataCell/* and geoDataCell/* components to visualize and edit the new "Geo" standard introduced by Announcements. This standard embeds WKT in the Data itself and allows multiple geometry to be inserted in the model
- Fix Toolbox visualization (Firefox) and height(chrome) when the sibling div (the content) was not covering the full page height
- Updated mapView component to be able to handle geometries and position coming from different data properties and format. For example Hicking have the gemotry encoded as reference to GeoShape entity, while Announcement have the full WKT geometry in the data but no Longitude/Latitude)
- Extended the arrayMapping capability in the **sharedView.ts** and **listView.ts** to allow us to handle array of objects by defining foreach nested object how to render each single field without the need of creating a new custom component and created a custom container EditNestedArrayCell which renders the inner array as table/tab. Recursive nesting is supported. Example:

```typescript
{
  title: 'Roads',
  component: CellComponent.EditNestedArrayCell,
  arrayMapping: {
    targetPropertyName: 'roads',
    pathToParent: 'RoadInfo',
    properties: [
      {
        title: 'Road Name',
        component: CellComponent.StringCell,
        objectMapping: { text: 'Name.{language}' }
      },
      {
        title: 'GPS Location',
        component: CellComponent.EditGpsInfoCell,
        objectMapping: {
          latitude: 'GpsInfo.Latitude',
          longitude: 'GpsInfo.Longitude'
        }
      },
      {
        title: 'Tags',
        component: CellComponent.ArrayCell,
        objectMapping: { items: 'Tags' }
      }
    ]
  }
}
```

### DIFF
--- a/databrowser/package-lock.json
+++ b/databrowser/package-lock.json
@@ -31,6 +31,8 @@
         "keycloak-js": "^26.2.0",
         "maplibre-gl": "^5.6.1",
         "maplibre-gl-draw": "^1.6.9",
+        "monaco-editor": "^0.54.0",
+        "monaco-editor-vue3": "^1.0.4",
         "pinia": "^3.0.3",
         "ramda": "^0.30.1",
         "swagger-client": "^3.34.3",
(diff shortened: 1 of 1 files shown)


gappc:
@Luscha thank you for your contribution, that's a very ambitious PR ;)

@Luscha @sseppi @RudiThoeni: unfortunately, it touches lots of the project's fundamentals (dependencies, code and UI / layout) and I'm sorry, but I cannot do a code review of such a huge PR with confidence. What were the requirements and reasons for the massive changes in the first place?

As a first step, would it be possible that you split the PR in smaller pieces, so it becomes more obvious what you're doing and how it is interconnected with the rest of the project? I see e.g. changes in how the UI config works, how data updates are merged and propagated and many more non-trivial stuff.

 Just, before you start splitting the PR, maybe lets discuss how we could tackle this PR?

gappc:
@sseppi yes that sounds like a good plan. If it's ok for you and @Luscha I'd start with the review as soon as the smaller PRs are here, otherwise it's almost impossible to understand what the impact of the changes would be.

gappc:
@Luscha thanks for the clarifications

As far as I understand, you're implementing some new components that don't need changes to the underlying config framework (`EditableWktMap`?), and then you implement a new generic component that requires some modifications (`EditObjectAsListCell`). If that's correct, I'd suggest to split the PR along these lines, while keeping an eye on PR sizes and feature complexity.

For the generic component, after reading your comments and looking at the proposed changes again, I think I understand the changes to the config framework better now and it makes sense. However, these changes will have an impact on many parts of the Data Browser, so we need to be very careful here. It would be great if you could provide a dedicated PR that focuses on these changes only, so we can review them in isolation.

gappc:
@sseppi I think this PR can be closed, it was split into two different other PRs that got already merged:

- PR #817
- PR #819

PR #811, opened 2025-12-05
Features: design review json editor

dallefeste:
This PR introduces the design review changes, the JSON editor and other minors.
Please note:

The branch was developed on my fork (dallefeste/it.bz.opendatahub.databrowser).
The works are still in progress regarding the push and sync requests and the attributes settings in details view.

### DIFF
--- a/databrowser/package-lock.json
+++ b/databrowser/package-lock.json
@@ -7120,9 +7120,9 @@
       }
     },
     "node_modules/js-yaml": {
-      "version": "4.1.0",
-      "resolved": "https://registry.npmjs.org/js-yaml/-/js-yaml-4.1.0.tgz",
-      "integrity": "sha512-wpxZs9NoxZaJESJGIZTyDEaYpl0FKSA+FB9aJiyemKhMwkxQg63h4T1KJgUGHpTqPDNRcmmYLugrRjJlBtWvRA==",
+      "version": "4.1.1",
+      "resolved": "https://registry.npmjs.org/js-yaml/-/js-yaml-4.1.1.tgz",
+      "integrity": "sha512-qQKT4zQxXl8lLwBtHMWwaTcGfFOZviOJet3Oy/xmGk2gZH677CJM9EvtfdSkgWcATZhj/55JZ0rmy3myCT5lsA==",
       "license": "MIT",
       "dependencies": {
         "argparse": "^2.0.1"
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -36,9 +36,9 @@ jobs:
       VITE_APP_KEYCLOAK_REALM: "noi"
       VITE_APP_KEYCLOAK_CLIENT_ID: "it.bz.opendatahub.databrowser"
       VITE_APP_KEYCLOAK_REDIRECT_URI: "https://databrowser.opendatahub.testingmachine.eu/silent-check-sso.html"
-      VITE_APP_IMAGE_UPLOAD_URL: "https://api.tourism.testingmachine.eu/v1/FileUpload/Image"
-      VITE_APP_FILE_UPLOAD_URL: "https://api.tourism.testingmachine.eu/v1/FileUpload/"
-      VITE_APP_ODH_LOOKUP_BASE_URL: "https://api.tourism.testingmachine.eu"
+      VITE_APP_IMAGE_UPLOAD_URL: "https://tourism.api.opendatahub.testingmachine.eu/v1/FileUpload/Image"
+      VITE_APP_FILE_UPLOAD_URL: "https://tourism.api.opendatahub.testingmachine.eu/v1/FileUpload/"
+      VITE_APP_ODH_LOOKUP_BASE_URL: "https://tourism.api.opendatahub.testingmachine.eu"
       VITE_APP_ENV_BADGE: "TESTING"
       VITE_APP_META_ROBOTS: "<meta name=\"robots\" content=\"noindex, nofollow\" />"
     steps:
@@ -135,9 +135,9 @@ jobs:
       VITE_APP_KEYCLOAK_REALM: "noi"
       VITE_APP_KEYCLOAK_CLIENT_ID: "it.bz.opendatahub.databrowser"
       VITE_APP_KEYCLOAK_REDIRECT_URI: "https://databrowser.opendatahub.com/silent-check-sso.html"
-      VITE_APP_IMAGE_UPLOAD_URL: "https://tourism.opendatahub.com/v1/FileUpload/Image"
-      VITE_APP_FILE_UPLOAD_URL: "https://tourism.opendatahub.com/v1/FileUpload"
-      VITE_APP_ODH_LOOKUP_BASE_URL: "https://tourism.opendatahub.com"
+      VITE_APP_IMAGE_UPLOAD_URL: "https://tourism.api.opendatahub.com/v1/FileUpload/Image"
+      VITE_APP_FILE_UPLOAD_URL: "https://tourism.api.opendatahub.com/v1/FileUpload"
+      VITE_APP_ODH_LOOKUP_BASE_URL: "https://tourism.api.opendatahub.com"
       VITE_APP_META_ROBOTS: ""
     steps:
       - name: Checkout source code
--- a/databrowser/public/robots.txt
+++ b/databrowser/public/robots.txt
@@ -1,0 +1,9 @@
+# SPDX-FileCopyrightText: NOI Techpark <digital@noi.bz.it>
+#
+# SPDX-License-Identifier: CC0-1.0
+
+User-agent: *
+
+Disallow: /*?*tagfilter=
+Disallow: /*?*source=
+Disallow: /*?*language=
--- a/databrowser/src/App.vue
+++ b/databrowser/src/App.vue
@@ -8,6 +8,13 @@ SPDX-License-Identifier: AGPL-3.0-or-later
   <router-view></router-view>
 </template>
 
+<script lang="ts" setup>
+import { useCanonicalHead } from './domain/contentpage/head/canonicalHead';
+
+// Head injection: set canonical URL
+useCanonicalHead();
+</script>
+
 <style>
 @font-face {
   font-family: 'SourceSansPro';
--- a/databrowser/src/components/checkbox/CheckboxCustomHomePage.vue
+++ b/databrowser/src/components/checkbox/CheckboxCustomHomePage.vue
@@ -1,0 +1,44 @@
+<!--
+SPDX-FileCopyrightText: NOI Techpark <digital@noi.bz.it>
+
+SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+
+<template>
+  <CheckboxCustom
+    v-model="checked"
+    class="mr-2"
+    :filter-key="filterKey"
+    :label="label"
+  />
+</template>
+
+<script setup lang="ts">
+import { computed } from 'vue';
+import CheckboxCustom from './CheckboxCustom.vue';
+
+const emit = defineEmits(['click', 'update:modelValue']);
+
+const props = withDefaults(
+  defineProps<{
+    filterKey: string;
+    filterLabel: string;
+    modelValue: boolean;
+    label?: string;
+    filterSelected?: { key: string; value: string }[];
+  }>(),
+  {
+    label: undefined,
+    filterSelected: () => [],
+  }
+);
+const checked = computed({
+  get: () =>
+    props.filterSelected.some(
+      ({ key, value }) =>
+        props.filterLabel === value?.toLowerCase() &&
+        props.filterKey === key?.toLowerCase()
+    ) || props.modelValue,
+  set: (value) => emit('update:modelValue', value),
+});
+</script>
--- a/databrowser/src/components/toggle/ToggleCustomHomePage.vue
+++ b/databrowser/src/components/toggle/ToggleCustomHomePage.vue
@@ -1,0 +1,46 @@
+<!--
+SPDX-FileCopyrightText: NOI Techpark <digital@noi.bz.it>
+
+SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+
+<template>
+  <ToggleCustom v-model="turnedOn" />
+</template>
+
+<script setup lang="ts">
+import { computed, watch } from 'vue';
+import ToggleCustom from './ToggleCustom.vue';
+
+const emit = defineEmits(['update:modelValue']);
+
+const props = withDefaults(
+  defineProps<{
+    modelValue?: boolean;
+    disabled?: boolean;
+    filterKey?: string;
+    filterSelected?: { key: string; value: string }[];
+  }>(),
+  {
+    modelValue: true,
+    disabled: false,
+    filterKey: '',
+    filterSelected: () => [],
+  }
+);
+
+const turnedOn = computed({
+  get: () => props.modelValue,
+  set: (value) => emit('update:modelValue', value),
+});
+
+watch(
+  () => props.filterSelected,
+  (newFilterSelected) => {
+    turnedOn.value = newFilterSelected.some(
+      (filter) => filter.key === props.filterKey
+    );
+  },
+  { immediate: true }
+);
+</script>
--- a/databrowser/src/config/builder/tourism/additionalProperties.ts
+++ b/databrowser/src/config/builder/tourism/additionalProperties.ts
@@ -2,9 +2,15 @@
 //
 // SPDX-License-Identifier: AGPL-3.0-or-later
 
-import { CellComponent } from '../../../domain/cellComponents/types';
-import { EditElements } from '../../../domain/datasets/config/types';
-import { echargingdataCategory } from './echargingdata';
+import { CellComponent } from '@/domain/cellComponents/types.ts';
+import { EditElements } from '@/domain/datasets/config/types.ts';
+import { echargingdataCategory } from './additionalproperties/echargingdata';
+import { poiagedataCategory } from './additionalproperties/poiagedata';
+import { activityltsdataCategory } from './additionalproperties/activityltsdata';
+import { poiltsdataCategory } from './additionalproperties/poiltsdata';
+import { gastronomyltsdataCategory } from './additionalproperties/gastronomyltsdata';
+import { suedtirolweincompanydataCategory } from './additionalproperties/suedtirolweincompanydata';
+import { siagmuseumdataCategory } from './additionalproperties/siagmuseumdata';
 
 export const additionalPropertiesCategory = (): EditElements => ({
   name: 'Additional properties',
@@ -27,6 +33,30 @@ export const additionalPropertiesCategory = (): EditElements => ({
     {
       objectPath: 'AdditionalProperties.EchargingDataProperties',
       elements: echargingdataCategory(),
+    },
+    {
+      objectPath: 'AdditionalProperties.PoiAgeDataProperties',
+      elements: poiagedataCategory(),
+    },
+    {
+      objectPath: 'AdditionalProperties.ActivityLtsDataProperties',
+      elements: activityltsdataCategory(),
+    },
+    {
+      objectPath: 'AdditionalProperties.PoiLtsDataProperties',
+      elements: poiltsdataCategory(),
+    },
+    {
+      objectPath: 'AdditionalProperties.GastronomyLtsDataProperties',
+      elements: gastronomyltsdataCategory(),
+    },
+    {
+      objectPath: 'AdditionalProperties.SuedtirolWeinCompanyDataProperties',
+      elements: suedtirolweincompanydataCategory(),
+    },
+    {
+      objectPath: 'AdditionalProperties.SiagMuseumDataProperties',
+      elements: siagmuseumdataCategory(),
     },
   ],
 });
--- a/databrowser/src/config/builder/tourism/additionalproperties/activityltsdata.ts
+++ b/databrowser/src/config/builder/tourism/additionalproperties/activityltsdata.ts
@@ -1,0 +1,29 @@
+// SPDX-FileCopyrightText: NOI Techpark <digital@noi.bz.it>
+//
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+import { CellComponent } from '@/domain/cellComponents/types';
+import { DetailElements } from '@/domain/datasets/config/types';
+
+export const activityltsdataCategory = (
+  options = { visible: false }
+): DetailElements => ({
+  name: 'Activity Properties',
+  slug: 'activityltsdata',
+  visible: options.visible,
+  subcategories: [
+    {
+      name: '',
+      properties: [
+        {
+          title: 'Altitude Difference',
+          component: CellComponent.StringCell,
+          objectMapping: {
+            text: 'AdditionalProperties.ActivityLtsDataProperties.AltitudeDifference',
+          },
+          params: { type: 'number' },
+        },
+      ],
+    },
+  ],
+});
--- a/databrowser/src/config/builder/tourism/additionalproperties/echargingdata.ts
+++ b/databrowser/src/config/builder/tourism/additionalproperties/echargingdata.ts
@@ -2,9 +2,9 @@
 //
 // SPDX-License-Identifier: AGPL-3.0-or-later
 
-import { CellComponent } from '../../../domain/cellComponents/types';
-import { DetailElements } from '../../../domain/datasets/config/types';
-import { withOdhBaseUrl } from '../../utils';
+import { CellComponent } from '@/domain/cellComponents/types';
+import { DetailElements } from '@/domain/datasets/config/types';
+import { withOdhBaseUrl } from '../../../utils';
 
 export const echargingdataCategory = (
   options = { visible: false }
--- a/databrowser/src/config/builder/tourism/additionalproperties/gastronomyltsdata.ts
+++ b/databrowser/src/config/builder/tourism/additionalproperties/gastronomyltsdata.ts
@@ -1,0 +1,29 @@
+// SPDX-FileCopyrightText: NOI Techpark <digital@noi.bz.it>
+//
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+import { CellComponent } from '@/domain/cellComponents/types';
+import { DetailElements } from '@/domain/datasets/config/types';
+
+export const gastronomyltsdataCategory = (
+  options = { visible: false }
+): DetailElements => ({
+  name: 'Gastronomy Properties',
+  slug: 'gastronomyltsdata',
+  visible: options.visible,
+  subcategories: [
+    {
+      name: '',
+      properties: [
+        {
+          title: 'Max SeatingCapacity',
+          component: CellComponent.StringCell,
+          objectMapping: {
+            text: 'AdditionalProperties.GastronomyLtsDataProperties.MaxSeatingCapacity',
+          },
+          params: { type: 'number' },
+        },
+      ],
+    },
+  ],
+});
--- a/databrowser/src/config/builder/tourism/additionalproperties/poiagedata.ts
+++ b/databrowser/src/config/builder/tourism/additionalproperties/poiagedata.ts
@@ -1,0 +1,37 @@
+// SPDX-FileCopyrightText: NOI Techpark <digital@noi.bz.it>
+//
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+import { CellComponent } from '@/domain/cellComponents/types';
+import { DetailElements } from '@/domain/datasets/config/types';
+
+export const poiagedataCategory = (
+  options = { visible: false }
+): DetailElements => ({
+  name: 'Age Properties',
+  slug: 'poiagedata',
+  visible: options.visible,
+  subcategories: [
+    {
+      name: '',
+      properties: [
+        {
+          title: 'Age From',
+          component: CellComponent.StringCell,
+          objectMapping: {
+            text: 'AdditionalProperties.PoiAgeDataProperties.AgeFrom',
+          },
+          params: { type: 'number' },
+        },
+        {
+          title: 'Age To',
+          component: CellComponent.StringCell,
+          objectMapping: {
+            text: 'AdditionalProperties.PoiAgeDataProperties.AgeTo',
+          },
+          params: { type: 'number' },
+        },
+      ],
+    },
+  ],
+});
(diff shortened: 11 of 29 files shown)


databrowser/src/locales/en.json
```diff
@@ -178,25 +193,96 @@
...
       "viewEmbeddedConfig": "Embedded view config",
       "viewGeneratedConfig": "Generated view config",
+      "viewUserShortLabel": "Custom",
+      "viewEmbeddedShortLabel": "Default",
+      "viewGeneratedShortLabel": "Raw",
+      "showDeprecatedFields": "Show deprecated fields",
+      "viewUserConfig": "User view config",
       "tableSettings": "Table settings",
-      "showDeprecatedFields": "Show deprecated fields"
+      "alert":{
```
gappc: Missing space between `"alert"` and `:` indicates, that prettier rules were not applied (see also general remark on PR). We should define a strategy on how to handle formatting issues.

databrowser/src/components/download/DownloadMenu.vue
```diff
@@ -67,22 +67,34 @@ SPDX-License-Identifier: AGPL-3.0-or-later
...
+                      :class="download.error ? 'w-1/2' : 'w-full'"
+                      :title="download.name"
+                  >
+                    {{ download.name }}
+                  </span>
+                  <!-- Download error -->
+                  <span
+                      v-if="download.error"
+                      class="ml-2 w-1/2 shrink-0 truncate text-hint-error cursor-copy"
+                      :title="download.error"
```
gappc: The `title` attribute is set on this `span`, in addition there is the `SimpleTooltip` component below, and both show the same content. As result, when you hover over an error, you'll see two tooltips. Which one is shown depends if you move the mouse over the error message or the retry button. I'd opt for one solution (see also other comment in this file regarding tooltips)

databrowser/src/domain/datasets/ui/toolBox/export/DownloadCard.vue
```diff
@@ -126,3 +152,9 @@ const download = () => {
   }
 };
 </script>
+
+<style scoped>
+.disabled{
+  opacity: 0.5;
+}
+</style>
```
gappc: Please do not use custom style definitions if not absolutely necessary, it makes styling and style updates harder and, in the worst case, produces a dedicated css file and `link` section in the DOM like the one below (copied from production build of this PR):

```html
<link rel="stylesheet" crossorigin="" href="/assets/ColumnConfigurationNavigationGuardHandler-DEEvVKjm.css">
```

Use [tailwind custom styles instead](https://tailwindcss.com/docs/adding-custom-styles)

databrowser/src/config/builder/tourism/syncdataconfig.ts
```diff
@@ -2,11 +2,12 @@
 //
 // SPDX-License-Identifier: AGPL-3.0-or-later
 
-import { CellComponent } from '../../../domain/cellComponents/types';
+import { CellComponent } from '@/domain/cellComponents/types.ts';
```
gappc: Please remove the `.ts` file extension

databrowser/src/components/select/SelectWithIconButton.vue
gappc: This component is almost the same as `SelectCustom.vue` (note that `SelectCustom.vue` was updated recently). Would it be possible to merge the two components?

databrowser/src/config/builder/tourism/pushconfig.ts
```diff
@@ -7,9 +7,10 @@ import {
   DetailElements,
   PropertyConfig,
 } from '../../../domain/datasets/config/types';
+import {DEFAULT_DATE_TIME_FORMAT} from "@/config/utils.ts";
 
 export const pushconfigCell = (): PropertyConfig => ({
-  title: '',
+  title: 'asasd',
```
gappc: Are you sure that this is the correct title? ;)

databrowser/src/locales/en.json
gappc: I see "Unusual line terminators" warning pop up in VSCode when I open this file (running on Linux)

Image

Could you please take a look at it?

databrowser/src/domain/datasets/ui/jsonView/JsonView.vue
```diff
@@ -0,0 +1,213 @@
...
+//     try {
+//       const parsed = JSON.parse(val || '{}');
+//       editStore.setCurrent(parsed);
+//       hasCurrentJsonErrors.value = false;
+//     } catch (e) {
+//       hasCurrentJsonErrors.value = true;
+//       console.warn('Invalid JSON in editor:', e);
+//     }
+//   },
+// });
```
gappc: Please remove commented code ;)

databrowser/src/domain/datasets/ui/common/editor/DiffChanges.vue
```diff
@@ -0,0 +1,63 @@
...
+  while (cells.length < 5) cells.push('gray');
+  return cells;
+});
+</script>
+<style scoped>
+/*noinspection CssUnusedSymbol*/
+.diff-square {
+  margin: 0 1px;
+}
+</style>
```
gappc: It seems that this style is used only in the [DiffChangesSquare.vue](https://github.com/dallefeste/opendatahub-databrowser/blob/61b8365b43fbf85592c8a0dc73165b987e4eb1e4/databrowser/src/domain/datasets/ui/common/editor/DiffChangesSquare.vue) component. If that's correct, then please remove this style definition here

databrowser/src/domain/datasets/ui/toolBox/ToolBox.vue
```diff
@@ -5,167 +5,109 @@ SPDX-License-Identifier: AGPL-3.0-or-later
...
       <Transition
         @after-enter="
           (el: Element) => (el.classList.value = 'block h-full w-full')
         "
         @before-leave="(el: Element) => (el.classList.value = 'hidden')"
       >
-        <div v-if="toolBoxStore.visible" :class="{ hidden: !mdAndLarger }">
-          <ContentAlignmentX class="h-full">
+        <div v-if="activeSection" :class="{ hidden: !mdAndLarger }">
+          <ContentAlignmentX class="h-full" :noPadding="true" >
```
gappc: I'd rather use a `div` here or create a dedicated `ToolboxAlignmentX` component (if reused several times), instead of modifying the `ContentAlignmentX` class just for this special purpose. Having `ContentAlignmentX` reused here makes the code harder to understand. My thoughts where: "why does `ContentAlignmentX` now support a `noPadding` prop? => ah, it's because it is reused once somewhere else where it is necessary to have no padding..."

Not reusing `ContentAlignmentX` would simplify the code

databrowser/src/domain/datasets/ui/toolBox/ToolBoxSection.vue
```diff
@@ -0,0 +1,19 @@
...
+</template>
+
+<script setup lang="ts">
+import type { Component } from 'vue';
+
+defineProps<{
+  sectionKey: string; // meglio evitare `key` perché è una parola speciale di Vue
+  title: string;
+  iconComponent?: Component;
+  info?: string;
```
gappc: This prop is not used by [ToolBox.vue](https://github.com/dallefeste/opendatahub-databrowser/blob/61b8365b43fbf85592c8a0dc73165b987e4eb1e4/databrowser/src/domain/datasets/ui/toolBox/ToolBox.vue), so I would remove it

databrowser/src/domain/datasets/ui/toolBox/ToolBox.vue
```diff
@@ -5,167 +5,109 @@ SPDX-License-Identifier: AGPL-3.0-or-later
...
-    @click="toolBoxStore.toggleToolboxVisibility(!toolBoxStore.visible)"
-  >
-    <IconAdd class="mr-2 size-5" />
-    <span>Toolbox</span>
-  </ButtonCustom>
 
   <!-- ToolBox content -->
   <div
-    class="absolute top-0 z-10 flex h-full flex-col overflow-x-auto bg-gray-50 transition-all md:relative"
+    class="absolute top-0 z-[10] flex h-full flex-col overflow-x-auto bg-gray-50 transition-all md:relative"
```
gappc: No need to add brackets to `z-10`, that class is already supported by tailwind

databrowser/src/domain/datasets/ui/tableView/toolBox/UserTableSettingsToolBoxPanel.vue
gappc: I think we don't need this component anymore, since the toolbox layout changed with the new design. If we remove this component, we could just use the `ColumnConfigurator` component in `TableViewToolBox.vue` and `DetailToolBox.vue` (although at the moment it makes no sense to allow column configuration in the `DetailToolBox.vue`, because there we have no table to configure in the detail view)

databrowser/src/domain/datasets/ui/detailView/toolBox/DetailToolBox.vue
gappc: Maybe this component should be renamed to `DetailViewToolBox`, because the corresponding component int the table view is named `TableViewToolBox`

databrowser/src/domain/datasets/ui/toolBox/export/ExportDatasetsToolBoxPanel.vue
```diff
@@ -5,102 +5,112 @@ SPDX-License-Identifier: AGPL-3.0-or-later
...
-
-    <ToolBoxSectionLabel v-if="!!referencesUrls">{{
-      t('datasets.toolBox.exportDatasets.sectionReferencesData')
-    }}</ToolBoxSectionLabel>
-    <ToolBoxCard
-      v-for="referenceUrl in referencesUrls || []"
-      :key="getReferenceKey(referenceUrl)"
+          <span>{{ t('datasets.toolBox.exportDatasets.apiDatasets.button') }}</span>
+        </ButtonCustom>
+      </ToolBoxCardHeaderButton>
```
gappc: Caution, you're putting here a button (`ButtonCustom`) inside a button (`ToolBoxCardHeaderButton`). I think you can remove the surrounding `ToolBoxCardHeaderButton`

databrowser/src/domain/datasets/ui/toolBox/export/ExportDatasetsToolBoxPanel.vue
```diff
@@ -5,102 +5,112 @@ SPDX-License-Identifier: AGPL-3.0-or-later
...
+      </span>
+      <ToolBoxCardHeaderButton
           :aria-label="t('datasets.toolBox.exportDatasets.apiDatasets.iconAlt')"
-          @icon-click="copyUrl"
+      >
+        <ButtonCustom
+            class="mt-4 flex items-center gap-3 px-2 py-1.5 text-sm"
+            :size="Size.xs"
+            :aria-label="t('datasets.toolBox.exportDatasets.download.iconAlt')"
+            @click="copyUrl"
```
gappc: Copying the URL is broken, it puts `[object PointerEvent]` into the clipboard

databrowser/src/domain/datasets/ui/toolBox/export/ExportDatasetsToolBoxPanel.vue
```diff
@@ -145,12 +156,16 @@ const onCopyReference = (url: string) => {
   copyReferenceUrl(url);
 };
 
-const { copy: copyReferenceUrl, copied: copiedReferenceUrl } = useClipboard();
+const { copy: copyReferenceUrl } = useClipboard();
```
gappc: This mechanism doesn't reset the icon from `IconCheck` to `IconCopy` after some time, like it is the case with `copyUrl` / `copiedUrl`. Note, that this bug exists also in the old code, it would be nice if we could fix it in this iteration

databrowser/src/domain/datasets/ui/toolBox/export/ExportDatasetsToolBoxPanel.vue
gappc: General advise for Vue components: a Vue component should have a single root node (e.g. `div`). Although Vue supports having many elements in a components root, it can be problematic, e.g. if you want to apply a class onto the component from its parent component, it may not work as expected and you'll get warnings in the console. See also https://vuejs.org/guide/components/attrs.html#attribute-inheritance-on-multiple-root-nodes and https://vuejs.org/guide/components/attrs.html#attribute-inheritance-on-multiple-root-nodes

databrowser/src/domain/datasets/ui/toolBox/ToolBoxCardHeaderButton.vue
gappc: I'm not sure if we need this component anymore. It is used 3 times any only in `ExportDatasetsToolBoxPanel.vue`, but two of that usages should probably be removed, because they result in two nested `button` elements. The last usage could be replaced, thus simplifying the code

databrowser/src/domain/datasets/ui/tableView/toolBox/SearchAndFilterToolBoxPanel.vue
```diff
@@ -5,110 +5,113 @@ SPDX-License-Identifier: AGPL-3.0-or-later
...
-        <ResetAllFilters @reset-all-filters="removeAllFilters" />
-      </ToolBoxCardHeader>
 
-      <ToolBoxCardBody
-        v-for="(filter, index) in tableFilters"
-        :key="`${filter.propertyPath}-${index}`"
-        class="flex flex-col gap-2"
+  <ToolBoxSectionLabel>
+      <span class="mr-4">
+        <span class="text-black">{{props.pagination.pageCount}}</span> <span class="text-dialog">{{ t('datasets.listView.recordsOutOf') }}</span> <span class="text-black">{{ props.pagination.totalItems }}</span>
```
gappc: No need to write `props. ...` in a Vue template, you can just write e.g. `pagination.pageCount`. The Vue compiler takes care of the rest

gappc (commented):
Pheww, that was a huge one. @dallefeste please let me know if you need more info

gappc:
Here a few other remarks:

- we should keep custom `style` blocks in `.vue` files at a minimum. Using Tailwind CSS utility classes is preferred. Only use custom styles when absolutely necessary.
- we should try to keep code format according to prettier and eslint rules
- the toolbox width differs between table view and detail view (try to open the "Export" panel in both views to see the difference)
- the PR contains lots of imports with `.ts` suffixes that should be removed
- most `console.log `statements left in the code should be removed
- new SVG in components/svg/odh/todo/*.vue seem a bit strange, they use ids, inline styles, ...
- new SVG in components/svg/odh/*.vue should be cleaned up to remove fixed colors where possible, so that CSS can style them (e.g. `fill="#50742f"` should be `fill="currentColor"`)
- in raw view with edit and diff enabled, when changing the json, each keystroke lets the cursor jump to the start of the text area (e.g. <http://localhost:3000/dataset/raw/tourism/v1/Article/4A418EFC-9EBA-0D28-C0CD-EA289006239C> => click "Edit" and then "Diff" and then type something)
- colors are not used uniformly, e.g. some places use `text-green-400`, others use `text-green-500` (note that this is also a problem in the old code). maybe it makes sense to define a set of standard colors in the Tailwind config and use those consistently
- there are features that are not implemented at the moment, e.g. in the table view each row has a "Action" button and there are two entries in the dropdown that are always disabled ("Open in Analytisc" and "Open in Data Quality"). In my opinion they should not be shown at all if they are not implemented yet

gappc:
Some TODOs after the merge is done (would not do it in this PR to not make it bigger):

- [ ] use "@/components/" and similar anywhere instead of relative paths like "../../../components/"
- [ ] format all code according to prettier and eslint rules

databrowser/package-lock.json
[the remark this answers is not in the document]
dallefeste: ok

databrowser/src/components/download/DownloadMenu.vue
```diff
@@ -67,22 +67,34 @@ SPDX-License-Identifier: AGPL-3.0-or-later
...
-                <div class="flex grow flex-col overflow-auto">
-                  <span>{{ download.name }}</span>
-                  <span v-if="download.error" class="text-hint-error">
-                    {{ download.error }}
+                <div class="flex grow items-center overflow-hidden">
+                  <!-- Download name -->
+                  <span
+                      class="truncate"
+                      :class="download.error ? 'w-1/2' : 'w-full'"
+                      :title="download.name"
```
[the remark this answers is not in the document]
dallefeste: ok

sseppi:
Hi @gappc,

thank you for the detailed review. I try to provide a feedback for all points where you cited me.

> * it is not obvious how to navigate to a record detail, it seems that double-clicking a row is needed, which is not intuitive on a web application. @sseppi @Xeroxi91 @dallefeste maybe we should add a `Details` button to the `ACTIONS` dropdown menu?

The double click was requested by Patrick, but here I would ask for a feedbakc to @Xeroxi91, to optimize the user experience for the user, who wants to access the detail view. 

> ## Detail view
> * the `View` and `Attributes` buttons in the header are clickable, but it is not defined yet how they should work in the Detail view. @sseppi @Xeroxi91 maybe they could be disabled like the `Filters` button until we defined how they should work?

@gappc @Xeroxi91 @dallefeste  I think we can't disable the `View` and `Attributes` buttons. We agreed in tha last meeting that we use:

- `View` button to switch from the default to the raw view
- `Attributes` button to open the right toolbox where only the switcher that enables the deprecated and/or referenced fields removing all other configuration option that are suitable only for the table view.

> * changing the view to `Raw` does not work anymore. @seppi what is the expected behavior here? Maybe disabling the `View` button in the Detail view could be a solution until this is defined?

@gappc @Xeroxi91 
The swith to the Raw view is an important feature, widely used by the internal team and developers. We should fix the missing functionality of the raw view.

sseppi:
> @sseppi @dallefeste on a side note and not specific to this PR: I just saw, that the `prettier` guidelines for code formatting are not always enforced, some files do not adhere to the formatting rules, but that is true for older files (from before this PR) and some files from this PR. The main problem is, that we don't enforce code formatting on commit and this problem keeps popping up throughout most code changes and PRs - including mine ;)
> 
> Not having enforced code formatting means, that changes to some part of a file could change other not-related parts of that file if a) the file was not correctly formatted in the first place and b) the file was then formatted after the feature / bugfix implementation e.g. automatically during save in the IDE. In effect this means, that PRs suddenly contain changes, that have nothing to do with the feature / bugfix being implemented.
> 
> I suggest that we add some pre-commit hook that runs prettier on all changed code, what do you say? @sseppi if you're ok, I'll create a dedicated issue.

@gappc I agree, but I would do this once we finished the merging of the PRs and we have a stable version online. Can you please open a dedicated issue and assign it to me, so I can program this activitiy for the next iteration?

databrowser/src/config/builder/tourism/pushconfig.ts
dallefeste: yah, but maybe could be better Pushconfig :) Fixed.

databrowser/src/domain/datasets/ui/common/editor/DiffChanges.vue
```diff
@@ -0,0 +1,63 @@
...
+import { storeToRefs } from 'pinia';
+import { useSingleRecordLoad } from '@/domain/datasets/ui/common/load/useSingleRecordLoad.ts';
+import DiffChangesSquare from '@/domain/datasets/ui/common/editor/DiffChangesSquare.vue';
+
+const { data } = useSingleRecordLoad();
+
+const datasetViewStore = useDatasetViewStore();
+const { diffAdds, diffDels } = storeToRefs(datasetViewStore);
+
+const modifiedLineCount = computed(() => {
```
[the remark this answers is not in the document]
dallefeste: correct, i've fixed it

databrowser/src/domain/datasets/ui/common/editor/DiffChanges.vue
dallefeste: removed

databrowser/src/domain/datasets/ui/common/editor/useJsonEditorConfigurator.ts
```diff
@@ -0,0 +1,94 @@
...
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+import { ref, watch, type Ref, nextTick, computed, onUnmounted } from 'vue'
+import type * as monaco from 'monaco-editor'
+import { DiffEditMode } from '@/domain/datasets/view/types'
+
+type Params = {
+  diffEditMode: Ref<DiffEditMode>
+  isEditEnabled: Ref<boolean>
+  isDiffEditing?: Ref<boolean>
```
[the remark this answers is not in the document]
dallefeste: done

databrowser/src/domain/datasets/ui/header/DatasetHeaderConfigPopup.vue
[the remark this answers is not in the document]
dallefeste: yes, i've removed.

databrowser/src/domain/datasets/ui/tableView/filter/tableFilterStore.ts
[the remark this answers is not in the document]
dallefeste: ok that's right, removed multiple triggers

databrowser/src/components/dropdown/Dropdown.vue
```diff
@@ -23,7 +23,7 @@ SPDX-License-Identifier: AGPL-3.0-or-later
       <!-- Dropdown menu -->
       <div
         v-show="show"
-        class="absolute z-10 min-w-full border bg-white shadow-xl"
+        class="absolute z-[999] min-w-full border bg-white shadow-xl"
```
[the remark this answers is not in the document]
gappc: @dallefeste well, I'm not 100% sure but at least it looks like the change came from this PR:

- `main` branch of `noi-techpark` has `z-10`: https://github.com/noi-techpark/opendatahub-databrowser/blob/main/databrowser/src/components/dropdown/Dropdown.vue#L26
- `feature/add-table-column-config-for-user` branch of `noi-techpark` has `z-10`: https://github.com/noi-techpark/opendatahub-databrowser/blob/feature/add-table-column-config-for-user/databrowser/src/components/dropdown/Dropdown.vue#L26
- `main` branch of your fork `dallefeste` has `z-10`: https://github.com/dallefeste/opendatahub-databrowser/blob/main/databrowser/src/components/dropdown/Dropdown.vue#L26
- `features/design-review-json-editor` branch of your fork `dallefeste` (same as this PR) has `z-[999]`: https://github.com/dallefeste/opendatahub-databrowser/blob/features/design-review-json-editor/databrowser/src/components/dropdown/Dropdown.vue#L26

Nevertheless, could you please check if it works with `z-10` and fix it if it's ok? Having arbitrarily high `zIndex` values usually turn out to become problematic

databrowser/src/components/select/SelectWithIconButton.vue
dallefeste: super, merged and removed the SelectWithIconButton.vue

databrowser/src/domain/datasets/ui/common/editor/DiffChangesSquare.vue
```diff
@@ -0,0 +1,25 @@
+<!--
+SPDX-FileCopyrightText: NOI Techpark <digital@noi.bz.it>
+
+SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+
+<template>
+  <div
+    class="h-3 w-3 mx-[1px]"
```
gappc: `mx-[1px]` can be written as `mx-px`

databrowser/src/domain/datasets/ui/header/DatasetNavigation.vue
```diff
@@ -34,52 +34,174 @@ SPDX-License-Identifier: AGPL-3.0-or-later
...
+  }
+};
+
+
+const breakpoints = useBreakpoints(breakpointsTailwind);
+const mdAndSmaller = breakpoints.smaller('lg');
+watch(
+  [isRawEditing, mdAndSmaller],
+  ([edit, isSmall], [wasEdit]) => {
+    console.log('edit:', edit, 'wasEdit:', wasEdit, 'isSmall:', isSmall);
```
gappc: Please remove `console.log` statements

Xeroxi91:
@dallefeste thank you for the update.
@sseppi I try to provide a feedback for all points where you cited me.
 
>  1. Clicking the span copies the error to the clipboard, but there is no user feedback. @sseppi @Xeroxi91 should we add here some user feedback?
> 
Adding user feedback can definitely help users understand the error.
> 
>  2. it is not obvious how to navigate to a record detail, it seems that double-clicking a row is needed, which is not intuitive on a web application. @sseppi @Xeroxi91 @dallefeste maybe we should add a Details button to the ACTIONS dropdown menu?
> 
I agree, double-clicking is not intuitive, so either we simplify it with a single click or we should consider reintroducing the Details button.

databrowser/src/domain/datasets/ui/toolBox/ToolBox.vue
gappc: Are the custom styles at the end of this file still in use? If not, please remove them

databrowser/src/domain/datasets/ui/detailView/toolBox/DetailViewToolBox.vue
```diff
@@ -0,0 +1,65 @@
...
+SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+
+<template>
+  <ToolBox>
+    <ToolBoxSection
+      :sectionKey="ToolBoxSectionKey.EXPORTS"
+      :title="t('datasets.toolBox.exportDatasets.panelName')"
+      :iconComponent="IconDownload"
+      info="Scarica i dati in diversi formati"
```
gappc: Prop `info` is not used neither in `ToolboxSection.vue` nor in `ToolBox.vue`, please remove this prop

gappc:
@sseppi @mrabans @Xeroxi91 @dallefeste I've deployed the current version of this PR as of commit [0dcf06b](https://github.com/noi-techpark/opendatahub-databrowser/pull/811/commits/0dcf06b135e865fbf437427dfd029ef36927a76e) to https://5.databrowser.gappc.net, so you can try it out.

@dallefeste nice updates, thank you :+1: I think we're nearing the finish line, but some things are still open:

## Open issues from previous code review
- the branch can not be merged, because there are merge conflicts, please fix them
- table view warnings (please fix, because this a more serious problem where the types don't match, see also https://github.com/noi-techpark/opendatahub-databrowser/pull/811#issuecomment-3631289173):
  ```
  [Vue warn]: Invalid prop: type check failed for prop "data". Expected Object, got Number with value 24. at <TableLinks data=24 record-id=undefined show-edit=false  ... > 
  ```
- detail view / raw view (see also https://github.com/noi-techpark/opendatahub-databrowser/pull/811#issuecomment-3631289173):
  - raw view does not work anymore in detail view (note that it **does** work in table view)
- please remove unused dependencies in `package.json` (see https://github.com/noi-techpark/opendatahub-databrowser/pull/811#discussion_r2602693011)
- remove console.log statements (see https://github.com/noi-techpark/opendatahub-databrowser/pull/811#discussion_r2602799921)
- please do not use custom style definitions (see https://github.com/noi-techpark/opendatahub-databrowser/pull/811#discussion_r2603336556)
- please remove emtpy style definitions for code cleanup (see https://github.com/noi-techpark/opendatahub-databrowser/pull/811#discussion_r2605684674)
- please add a short description to `ToolboxSection.vue` to explain how this component works together with `Toolbox.vue`, it will save us and others a lot of headaches in the future (see https://github.com/noi-techpark/opendatahub-databrowser/pull/811#discussion_r2605832367)
- remove unused component prop `info` from `ToolboxSection.vue` (see https://github.com/noi-techpark/opendatahub-databrowser/pull/811#discussion_r2605927304)
- replace `z-[10]` class with `z-10` in `Toolbox.vue` (see https://github.com/noi-techpark/opendatahub-databrowser/pull/811#discussion_r2605946028)
- rename component `ToolboxSectionHeader.vue` to `ToolboxHeader.vue` for clarity (see https://github.com/noi-techpark/opendatahub-databrowser/pull/811#discussion_r2605961016)
- it would be nice if we could clean up the `withBgColor` stuff (see https://github.com/noi-techpark/opendatahub-databrowser/pull/811#discussion_r2606067510)
- fix broken link copy in export (see https://github.com/noi-techpark/opendatahub-databrowser/pull/811#discussion_r2606218582) and fix icon sizes (see https://github.com/noi-techpark/opendatahub-databrowser/pull/811#discussion_r2606230122)
- issue about icon reset on link copy is still open (see https://github.com/noi-techpark/opendatahub-databrowser/pull/811#discussion_r2606243240)
- make `TableLinks.vue` component reactive (see https://github.com/noi-techpark/opendatahub-databrowser/pull/811#discussion_r2606472960)
- remove `props...` to access component props in template (see https://github.com/noi-techpark/opendatahub-databrowser/pull/811#discussion_r2606716896)

## Current code review

Please see my latest code review feedback from today (2025-12-16) - mostly cleanup

## New bugs

- Diff stats and button stick together
  image
- bug in detail view when JSON is shown and Edit mode is activated => when the user changes the JSON, there is no hint that data changed and that it can be saved. steps to reproduce:
  - log in to the data browser
  - open the raw / json view of any record (e.g. http://localhost:3000/dataset/raw/tourism/v1/Accommodation/5CEA544EE34639034F07B79D4AEEB603)
  - activate edit mode by clicking on "Edit" button on the top right
  - change JSON in the editor below
  => situation as of this PR (commit 0dcf06b): the "Save" and "Cancel" buttons don't appear and the navigation guards don't execute when the user tries to navigate away from that page
  => expected: the "Save" and "Cancel" buttons do appear and the navigation guards show a popup indicating that there are unsaved changes

sseppi:
> > > * changing the view to `Raw` does not work anymore. @seppi what is the expected behavior here? Maybe disabling the `View` button in the Detail view could be a solution until this is defined?
> > 
> > 
> > @gappc @Xeroxi91 The swith to the Raw view is an important feature, widely used by the internal team and developers. We should fix the missing functionality of the raw view.
> 
> Regarding this point: my changes should not change the view handler functionality. I double-checked the behavior directly on the online version [(here),](https://databrowser.opendatahub.com/dataset/detail/tourism/v1/Accommodation/5CEA544EE34639034F07B79D4AEEB603_REDUCED?language=en) and the Raw view switch appears to be working in the same way.
> 
> Could you please clarify what the expected behavior is? Thanks.
> 
> image

@dallefeste OK, fine!

